PR #279, opened 2014-06-10
submitting pattern for a video element  background

mpiotrowicz:
not a super wide use case but could be useful nonetheless

### DIFF
--- a/patterns/index.html
+++ b/patterns/index.html
@@ -21,6 +21,7 @@ description: An accessible widget & pattern library
         <h3>Imagery</h3>
         <ul>
             <li><a href="http://codepen.io/grayghostvisuals/pen/eyvrK">SVG</a></li>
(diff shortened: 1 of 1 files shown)


grayghostvisuals:
Hey there @mpiotrowicz You wanna pull this in? Up to you to close or merge :)

mpiotrowicz:
Hooookay, I've added play/pause functionality as suggested in the above article. I've also added a caveat with a link to the article. I'm torn because overall this isn't always a great pattern for people to use so I don't want to endorse it necessarily, buuut, this is a way of implementing that pattern in a way that's much friendlier. If we get flack from anywhere, I'll happily remove

svinkle:
The icons should be reversed. Currently, while the video plays the 'play' icon is shown when it should be the 'pause' icon. When paused the 'play' icon should be displayed.

mpiotrowicz:
yikes good catch! fixed

davatron5000:
:+1:

Issue #310, opened 2015-01-12
Markdown version of a11y checklist


chris-dura:
Hello,

I work on a huge enterprise development team, and we make heavy use of GitHub issues, gists and wikis. In an effort to mitigate some of the headaches when managing all the Pull Requests from a gargantuan team, we've instituted a "PR template" that all contributors have to use when they open a PR. 

Basically, it's a boilerplate markdown file that has standard tasks the developer needs to complete before the PR is eligible to even be reviewed or considered a merge candidate. For example, things like "Unit tests pass", "Lint free", etc. Anyway, as the mouthpiece for a11y around the office, I'd like to start making basic accessibility one of those minimum requirements to get a PR merged.

I'm still pretty new to accessibility myself, but I've found your checklist seems like a good starting point. Obviously, a truly accessible project would need focused effort, but I think your checklist represents very simple things that _any_ developer could implement without being an "accessibility expert" and it would improve our project immeasurably and be an "easy win".

I think others might find it useful to have a markdown version of the checklist for a variety other reasons, but for me, I'd especially like a version that I could embed in Pull Requests. This way, it would save the dev's progress and serve as a running progress report on the PR, e.g. 10 tasks (3 completed, 7 remaining).

bartveneman:
What a great idea! I think I'll give that list a try myself. Thanks!

PR #311, opened 2015-01-12
Create checklist.md

chris-dura:
Create a markdown version of `checklist.html` so other teams can insert it into GitHub wikis, issues or even PRs.

Fixes https://github.com/a11yproject/a11yproject.com/issues/310

### DIFF
--- a/checklist.md
+++ b/checklist.md
@@ -1,0 +1,84 @@
+## Landmarks
+ARIA Landmark Roles are helpful landmarks that can be used by AT to navigate a website.
+
+- [ ] `<header role="banner">` 
+A region of the page that is site focused. Typically your global page header.
+
+- [ ] `<nav role="navigation">` 
+Contains navigational links.
+
+- [ ] `<main role="main">` 
+Focal content of document. Use only once.
+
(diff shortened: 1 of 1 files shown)


grayghostvisuals:
I wonder if we can alleviate the need to maintain two diff versions manually. Maybe somewhere near unicorn river is a Grunt task that can export our HTML to Markdown on builds?

grayghostvisuals:
Boom. [https://github.com/paazmaya/grunt-html2md](https://github.com/paazmaya/grunt-html2md)

joe-watkins:
@davatron5000 I agree managing two files isn't probably the way to go.

Per @grayghostvisuals automated suggestion I did a quick test locally with a Gulp HTML -> Markdown task and because the format of our checklist page is pretty unique, HTML-wise, the output probably wouldn't work for what @chris-dura is looking for. We are leaning on some CSS to manage the look and feel with absolute positioning.

I'm also wondering if the addition of the dependency of a task runner e.g. Gulp/Grunt isn't worth it for this feature addition?

This sounds like a nice-to-have that may not justify the level of effort or dependency needs.

PR #317, opened 2015-02-15
Added "Reframing Accessibility for the Web" post

AllThingsSmitty:
Excellent ALA article about changing how we think of accessibility.


davatron5000:
I really like the Accessibility Matrix part of the post. I think we should add it in. :+1: 

I read an [interesting counter-post to this](http://www.chemnitzer-14.de/accessibility-is-usability-in-context-of-disability/). I'm not saying we shouldn't block the ALA article, but it's a bit of an interesting discussion.

AllThingsSmitty:
@davatron5000 I hadn't seen that post; thanks for sharing. It's an interesting counter, definitely worthy of further discussion.

PR #321, opened 2015-02-23
Patch/upgrade bootstrap to 3.0

joe-watkins:
Here's the port to 3.0 for review.  If we decide to upgrade Bootstrap this should do the trick :)

Here are some notable changes.

1. Upgraded to Sass BS 3.0 - /sass/vendor/bootstrap /js/vendors/bootstrap
2. Made call to garlic.js local - one less external dep - kept jQuery external for caching benefits
3. Setup Sass folder/s to be more modular using partials
4. Updated markup from BS 2.0 to 3.0 specs
5. Added an 'active' state to main nav via js
6. Increased overall type size a few pixels for readability
7. Added a codekit config file for anybody that might use that tool


grayghostvisuals:
@joe-watkins So…we're not going to use bootstrap going forward making this is a wasted effort I'm sorry to say. It really breaks my heart to close this just so u know I'm not cold hearted :)

What would be really great is to get the site down to it's minimal form (markup) and set up a dir structure and update our dependency chains (maybe get gulp or grunt in the mix here). I think I'm gonna start this now just so we get some movement going forward.

Issue #320, opened 2015-02-22
Revisiting the footer


svinkle:
I love the footer, the concept and how it looks. The problem is all the images are adding a whopping 3Mb to each page load. Not sure what future plans are for the footer but this is a big problem.

My first instinct would be to implement a lazy-load feature for these images, only download to the client if the user reaches the footer area.

Any other thoughts?

grayghostvisuals:
@svinkle Agree with you 110%

joe-watkins:
I can handle this :)

PR #327, opened 2015-04-08
#320 - add lazyload to footer

joe-watkins:
Adding lazy load functionality to footer. Subsequent loads the footer is drawn from local storage. Anybody wanna take a peek?

### DIFF
--- a/js/scripts.js
+++ b/js/scripts.js
@@ -14,7 +14,35 @@ $(document).ready(function(){
 
     ui : {
 
+      lazyLoadFooter : function(target,offset){
+
+        var $target = $(target),
+            targetLoc = $target.offset().top,
+            footerLazyLoadTriggered = false;
+
+            if(!footerLazyLoadTriggered){
+
+              $(window).scroll(function(e){
+
+                var scrollPos = $(this).scrollTop(),
+                    offSet = offset,
+                    targetScroll = targetLoc - offSet;
+
+                    if(targetScroll <= scrollPos && !footerLazyLoadTriggered){
+                      Engine.ui.footerContributors(); // run custom footer
+                      footerLazyLoadTriggered = true;
+                    }
+
+              });
+
+            }else{
+              Engine.ui.footerContributors(); // run custom footer
+            }
+
+      }, // lazyLoadFooter()
+
       footerContributors : function(){
+
         function gitHubContributors(){
 
           var
@@ -260,7 +288,7 @@ $(document).ready(function(){
 
   } // Engine
 
-  Engine.ui.footerContributors();
+  Engine.ui.lazyLoadFooter("footer[role='contentinfo']",800);
   Engine.ui.footerCopyright();
   Engine.ui.toc();
 


js/scripts.js
```diff
@@ -14,7 +14,35 @@ $(document).ready(function(){
...
+        var $target = $(target),
+            targetLoc = $target.offset().top,
+            footerLazyLoadTriggered = false;
+
+            if(!footerLazyLoadTriggered){
+
+              $(window).scroll(function(e){
+
+                var scrollPos = $(this).scrollTop(),
+                    offSet = offset,
```
svinkle: Just stick to using `offset` instead of creating a new var.

js/scripts.js
```diff
@@ -14,7 +14,35 @@ $(document).ready(function(){
...
+
+                    if(targetScroll <= scrollPos && !footerLazyLoadTriggered){
+                      Engine.ui.footerContributors(); // run custom footer
+                      footerLazyLoadTriggered = true;
+                    }
+
+              });
+
+            }else{
+              Engine.ui.footerContributors(); // run custom footer
```
svinkle: Does this need to be here? Once the footer runs it doesn't need to run again.

js/scripts.js
```diff
@@ -14,7 +14,35 @@ $(document).ready(function(){
...
+
+        var $target = $(target),
+            targetLoc = $target.offset().top,
+            footerLazyLoadTriggered = false;
+
+            if(!footerLazyLoadTriggered){
+
+              $(window).scroll(function(e){
+
+                var scrollPos = $(this).scrollTop(),
```
svinkle: Should cache `$(window)` as `$window` and use it here as well instead of `$(this)` (since there is only one window object).

js/scripts.js
```diff
@@ -14,7 +14,35 @@ $(document).ready(function(){
...
+                    offSet = offset,
+                    targetScroll = targetLoc - offSet;
+
+                    if(targetScroll <= scrollPos && !footerLazyLoadTriggered){
+                      Engine.ui.footerContributors(); // run custom footer
+                      footerLazyLoadTriggered = true;
```
svinkle: Instead of this you could namespace your event, then remove it when the footer runs. For example, [:25](https://github.com/a11yproject/a11yproject.com/pull/327/files#diff-549b6b2bff82a0ad6730c0f1cd665a19R25) could be:

```javascript
$window.on('scroll.lazyLoadFooter', function(e){
```

Then here you could have

```javascript
$window.off('scroll.lazyLoadFooter');
```

This would remove the scroll event after the footer runs, cease polling on scroll.

svinkle:
This is really awesome. Exactly what I wanted to see happen. I tried this in my own setup and it works well. Few comments from me above.

joe-watkins:
@svinkle Thanks ! I'll work in your great JS suggestions.. thanks for sharing those.

joe-watkins:
Going to scratch this PR and do a new one and will include @svinkle 's JS fixes. Merge conflicts are nuts because I'm so far from master.

PR #328, opened 2015-04-10
Fix misspelling of the word "Accessibility"

rdmurphy:
I've looked at this page multiple times and I just now noticed it was misspelled. Whoops!


joe-watkins:
Good eye! :)

PR #358, opened 2015-09-09
Added Additional Reading to vestibular disorders page.

Aurif3x:
This article gives clear examples and helpful tips on how to use animation in ways that are less likely to make people sick. It's essentially an in-depth version of this page's "What should you consider?" blurb. I think it's worth linking to!


joe-watkins:
Gold! tnx ;)

PR #372, opened 2015-11-16
Updates old link on "MYTH: Accessibility is "blind people"" post

michaelck:
Noticed an old link on the "MYTH: Accessibility is "blind people""
post. It no longer resolves to the intended article. 

Located the original linked article and updated.


davatron5000:
:+1: Thanks and nice catch!

PR #374, opened 2015-12-01
Add practical ARIA examples to Further Reading Resources

coreybruyere:
Added Heydon's list of ARIA examples


joe-watkins:
tnx @coreybruyere good pick!

Issue #377, opened 2015-12-19
Article: Color Contrast


RachelRVasquez:
Here's another work in progress:

https://gist.github.com/RachelRVasquez/d1c1e99cfdd9780c5918

Tackling it from the article list under "Concepts": https://github.com/a11yproject/a11yproject.com/issues/12

Let me know if it sounds like it's going in the right direction so far. :-)

RachelRVasquez:
I think I'm just about done with this one - just not quite sure how I can wrap it up?

thierryk:
You could mention [high-contrast mode](https://developer.yahoo.com/blogs/ydn/techniques-high-contrast-friendly-icons-153038779.html) as few people are aware of this issue.

RachelRVasquez:
Added a link from W3C about high contrast style switchers and mentioned in the "What else can we do" list about providing tools to adjust to a lower/higher contrast. It sounds like it might be useful to write additional articles or checklists about techniques for high and low contrast tools.

thierryk:
@RachelRVasquez 

When it comes to high-contrast mode, it is more about techniques than tools, so I'm not sure the W3C link is of much help&mdash;why not linking to the article I mentioned earlier?

RachelRVasquez:
Added. :+1:

davatron5000:
:+1: This is great @RachelRVasquez! Do you need help article-izing this? Or do you want to make a pull request?

RachelRVasquez:
@davatron5000 I do need help actually. :-) I wasn't sure what the process was from this point onward. If you can tell me how to "article-ize" I'd like to learn how.

davatron5000:
Article-izing is maybe a bit difficult if you're not familiar with the Jekyll system. Here's the process in a nutshell:

1. Download the project in something like [GitHub Desktop](https://desktop.github.com/) or however you do `git`
1. Open the project folder, then open `_posts`
1. Duplicate `example_post.md` (https://github.com/a11yproject/a11yproject.com/blob/gh-pages/_posts/example-post.md) and retitle the file `YYYY-MM-DD-seo-title-slug.md`
1. Open the file in a text editor and paste your text below the `---` metadata
1. Fill out the metadata and set `published: true`
1. Save the file
1. "Commit" the file.
1. Use the "Pull Request" button to send a pull request.

[...]

PR #382, opened 2016-01-05
Color Contrast Article

RachelRVasquez:
This is the complete article from my gist on what color contrast is and
why we need it for web a11y

### DIFF
--- a/_posts/2015-01-05-what-is-color-contrast.md
+++ b/_posts/2015-01-05-what-is-color-contrast.md
@@ -1,0 +1,47 @@
+---
+layout: post
+title: What is Color Contrast?
+description: What is color contrast and why do we need it for web accessibility?
+author: Rachel R. Vasquez
+published: true
+categories:
+  - Basics
+---
+
+##Color Theory
+
+Colors from different segments of the color wheel are contrasting colors. In color theory, [complementary colors](https://en.wikipedia.org/wiki/Complementary_colors) (or "opposite" colors) that are directly across from one another on a basic color wheel provide maximum contrast. This means they stand out from one another the strongest compaired to being paired with other colors. Complementary colors are two colors that cancel one another out and produce a white/gray-scale color.
+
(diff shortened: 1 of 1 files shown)


RachelRVasquez:
For reference - this is linked to this article issue: https://github.com/a11yproject/a11yproject.com/issues/377#issuecomment-168840502

davatron5000:
LGTM. I think this is ready to go. Does anyone else want to QA this? @joe-watkins @mpiotrowicz? Otherwise I'll pull the trigger in a few minutes.

mpiotrowicz:
good content, I like the color theory background! Would it be appropriate to flesh out more examples of why good color contrast is important? Eg when faced with screen glare, being outside on a sunny day, anyone who's stared at a screen at work all day, etc. (though that might be better content for the impairment article)

joe-watkins:
Great work here.. I love it! Very readable :) To piggyback here I have a couple items:

**Edit:**
"There's a link to W3C's article at the end of this called "The Contrast Minimum" that explains how these ratios are calculated."

I'd love to see a meaningful link presented to the reader at this location rather than directing them to the end of the article.

**Suggestion:**
Also it may be worth pointing out that too much contrast can be an issue as well :) User's with Dyslexia can have trouble with Black on White: http://uxmovement.com/content/6-surprising-bad-practices-that-hurt-dyslexic-users/

davatron5000:
Cool. Let's merge this in and then iterations can be done in another issue and/or via the GitHub Editor.

davatron5000:
:clap: :clap: :clap: It's LIVE! http://a11yproject.com/posts/what-is-color-contrast/

Thanks @RachelRVasquez so much for hopping on board with this project. We all really appreciate the time, work, and new blood you've donated to this project. Look forward to the next one we get to merge in.

RachelRVasquez:
Yay thanks @davatron5000! I'm having a lot of fun and learning from you all as well.

@mpiotrowicz @joe-watkins These sound like great ideas for continuations. :-) I was having trouble deciding how deep to go into the subject so I tried to write it as a "primer" of sorts or an introduction. Then thinking maybe we can write other articles that go into depth of the different contrasts and visual impairments.

I like Joe's idea of editing the Contrast Minimum link so it's in the content versus at the end in "Further Reading". Wondering if maybe we can edit the description of the post so it's less repetitive as well?

PR #383, opened 2016-01-05
added What Input? to the list of scripts to remove outlines

FezVrasta:
I think "What Input?" is the best solution if a front-end developer needs to remove outlines from the elements of the page.

It allows you to style your elements in a specific manner only if the user is using a specific input method (mouse, touch, keyboard) and does not interfere with the click events.
Additionally, it's much more lightweight.

In the startup I work for we used it successfully and now most of the app is keyboard accessible.


davatron5000:
Nice find! I like it.

joe-watkins:
Hmm... this is a cool idea but makes me a little nervous.. spidey senses tingling. In a quick test you could see here how a user could be disoriented if this plugin is used how prescribed. http://www.screencast.com/t/VqUzNs3nZz

In this screencast I'm using both a mouse and a keyboard for navigating the interface found [on this Pen](http://codepen.io/joe-watkins/pen/jWBVGK/). You'll see the plugin remove the outline as I move the mouse outside the viewport as it bounces between keyboard and mouse assumption.

I still need to test on a laptop at home that has both touch and mouse... but perhaps we should make a note about using with extreme caution? Thoughts?

FezVrasta:
Didn't noticed this behavior, I don't think it's a big problem but I'll open an issue to see if it can be fixed

PR #389, opened 2016-01-19
Reordered links alphabetically

AllThingsSmitty:
Following the pattern that resources should be listed alphabetically
(https://github.com/a11yproject/a11yproject.com/blob/gh-pages/CONTRIBUTI
NG.md#adding-resources). Also, added quotes (“”) around the Read More
heading to better identify what the section represents. 👍🏼

### DIFF
--- a/patterns/index.html
+++ b/patterns/index.html
@@ -28,9 +28,9 @@ description: An accessible widget & pattern library
     <div class="span4">
         <h3>Accordions &amp; Tabs</h3>
         <ul>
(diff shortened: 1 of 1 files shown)


davatron5000:
@AllThingsSmitty Hey, this PR probably should have gone in pre-redesign sorry. The templates have kinda changed so it might be hard to successfully merge these together.

I've started thinking it might make sense to move Resources over to a `_data/resources.yml` file. That would be easier to maintain and alphabetize.

I think we should maybe close this and pursue that option. Thoughts?

AllThingsSmitty:
@davatron5000 sounds good to me. Thanks. :+1:

joe-watkins:
Patterns converted to `.yml` in 471

Issue #390, opened 2016-01-20
A11Y Simple Tabs without jQuery


BeyondHyper:
http://codepen.io/BeyondHyper/pen/xZXXzj

This is a tab solution based on Scott Vinkle's Simple Tabs (http://codepen.io/svinkle/pen/edmDF) removing the jQuery dependency and an enhanced presentation.

joe-watkins:
@BeyondHyper Nice work! Heya.. are you able to create a PR for the pattern? If not let me know and I'll do it.

BeyondHyper:
@joe-watkins - thanks! Yeah, I can do that.

svinkle:
@BeyondHyper, nice work! 

FYI I recently updated my pen to include a `focus` event listener along with the `click`. Reason behind this change is when someone using the VO Rotor would find one of the tabs under the Form Controls menu, the tab panel content would not get updated on click. This small change helps to make sure when a tab gets focus, by other means other than the arrow keys, the tab panel content loads the appropriate content. With this change you could also remove the `tabClickEvent()` calls from the `keyboardEvent()` function, but test first. :)

BeyondHyper:
@svinkle , thanks for the info! I don't think most people (including me) were/are thinking about voice over. I will take a look and adjust accordingly.

thierryk:
I think it is important that any example posted offers a proper fallback (i.e. with JS off). 
For example, in @BeyondHyper's last Pen the jump links do not seem to work without JS.

BeyondHyper:
@thierryk: Agreed! I just updated my anchors. It was a typo issue.

BeyondHyper:
@svinkle - wondering the reason for adding a click and focus listener since the item will gain focus when clicked and call the tabClick function? Also, in your current repo, when you click the tab, it is jumping to the anchor. Is this expected behavior?

svinkle:
@BeyondHyper, just fixed the jump-link issue by separating the `click` and `focus` events in to their own listener. Should be good now.

BeyondHyper:
@svinkle - I separated mine as well and am just calling the preventDefault method on click since the element will gain focus on click and fire the tabClick function.

thierryk:
I just created a pen for this ([Pure CSS Tab Panel](http://codepen.io/thierry/full/gPoWxj/)), not sure if it is worth adding it to the list though...

svinkle:
[...]

PR #391, opened 2016-01-20
Issue #390 - Added new tab pattern

BeyondHyper:


thierryk:
I think all these examples should include **links** in panels so we can see how _tabbing navigation_ works (as [#390](https://github.com/a11yproject/a11yproject.com/issues/390) does). This is because in most casses, tabbing through/past these panels is very confusing for users.

Back in the day, I created a solution that plugged 2 links at the end of each panel - one to reach the heading of the previous panel (if any) and one to reach the heading of the next panel (if any). This way, keyboard users can navigate **between** panels easily.

joe-watkins:
@thierryk That's a great idea! I forked @BeyondHyper 's pen and added links in the panels and it held up great for keyboard navigation.

@BeyondHyper you mentioned in original issue that your tab pattern was based on @svinkle 's tab pattern .. if it was more than loosely based perhaps we should consider giving credit or point to original Pen from your Pen?

BeyondHyper:
@joe-watkins - I can add the credit in the source code if you think that is better. I already provided credit in the pen description.

joe-watkins:
@BeyondHyper tnx! yes a code comment attribution would do the trick since folks are taken to the edit state over at CodePen. Great contribution!

BeyondHyper:
@joe-watkins thanks and all set! Added it to the js source.

BeyondHyper:
@joe-watkins and @thierryk - I made a couple of changes to the pen. I changed the set up logic so that it doesn't overwrite the tab panel id attribute and implemented @thierryk's idea of having next tab navigation links.

thierryk:
@BeyondHyper 

I think the next tab link is a great improvement for keyboard users but I think you should have:
   * a "next tab" link in the first panel
   * a "previous tab" link **and** a "next tab" link in the second panel
   * a "previous tab" link in the last panel

If you wanted to avoid visual clutter for pointing device users who don't have the need for these links, you could hide them (via `opacity`, off-screen, or else) and reveal them when they get focus (like we often do with skip links).

Makes sense?

BeyondHyper:
@thierryk - that makes sense. Thanks for the feedback!

PR #393, opened 2016-01-29
Fix spelling mistakes

pra85:


davatron5000:
Thanks and nice catch!

PR #401, opened 2016-02-22
Remove broken link

andkerel:
"[10 Easy Things All Web Developers Should Do](http://laneshill.me/2012/12/10-easy-things-all-web-developers-should-do-for-web-accessibility/)" currently not active


davatron5000:
Thanks but this and another URL were fixed by #409, so I took that one. Appreciate the help and anytime you see something out of date, make a PR and I'll try to be more punctual.

PR #404, opened 2016-03-16
Add slash at the end in permalinks of posts

ismamz:
Error 404 in links with slash at the end.
For example: https://github.com/a11yproject/a11yproject.com/blob/gh-pages/checklist.html#L104
Add slash at the end in permalinks of posts to ensure that work well in both cases.

(I edit CNAME but it was an error, sorry)

### DIFF
--- a/CNAME
+++ b/CNAME
@@ -1,1 +1,1 @@
-a11yproject.com
+a11yproject.github.io
--- a/CNAME
+++ b/CNAME
@@ -1,1 +1,1 @@
-a11yproject.github.io
+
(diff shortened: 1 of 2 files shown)


CNAME
```diff
@@ -1 +1 @@
-a11yproject.com
```
davatron5000: This is a mistake I think? We need the domain name in the CNAME so it resolves properly on the Github side of things.

ismamz: Yes, I made a mistake with `CNAME` file. But `_config.yml` I think that you need fix it.

ismamz: I created a new pull request: https://github.com/a11yproject/a11yproject.com/pull/405
Delete this: https://github.com/a11yproject/a11yproject.com/pull/404

PR #408, opened 2016-03-19
added blue glowing focus to linked images and btns

jeryj:
I added a .btn class for reuse (surprised I didn't do that originally...) and added a blue glowing outline that mimics chrome's focus state (the blue glow). This same blue glow is used for images that are children of links.

Two thoughts: 
1. Should the added focus be a blue glow? Is there a better default?
2. Is the blue glow enough of a contrast against our btn color? It's noticeable when the change happens when tabbing through, but not super visible. I'm concerned it'll be easy to lose track of where you are. Should the focus color be different? Other ideas on how to indicate focus in some kind of standardized way?


davatron5000:
I think something is better than nothing. I know Firefox and Edge default to dashed outlines. It's possible you'd need both `:focus` states (glow on a button, dashed on a link).

Maybe it's two placeholders that we can incorporate as necessary.

```scss
%glow {
  //stuff
}

%dash {
  //stuff
}

.post a:focus { @include %dash; }
.btn:focus { @include %glow; }
```

I'm also happy to roll what you have in an iterate from there. Something is better than nothing right now I think.

jeryj:
Agreed. A placeholder would be fine. I just used a %focus placeholder to override the default focus. Probably could have been more descriptive with %glow_outline and %dotted_outline. 

I'm planning to ask on the a11y slack channel to see if anyone has an opinion on what the best implementation should be.

j-mes:
This is slightly off topic: I'm very much interested in accessibility, and I've been following the repo for a while. Did someone say there's an a11y slack channel?! I'd be very interested in joining that and learning from other people who are a11y-focused.

jeryj:
Hi @j-mes! You can get an invite to the slack channel here: http://web-a11y.herokuapp.com/

jeryj:
I removed the tabindex values on <sections>, as they were unnecessary and I'm not sure why I added them in the first place...

joe-watkins:
@jeryj can you look into fixing the merge conflicts on this PR. The `tabindex` removal may be the culprit since patterns and resources has changed since your original PR.

I believe those `tabindex="0"` are on those containers to support jump links in right column. They could be `tabindex="-1"` to not create another tab stop.

jeryj:
@joe-watkins Sure! Thanks for pulling this out from the murky depth. I'd forgotten about it. I'll try to get this done this week and send it back. I think I have a lot of little tweaks throughout, so it would be good to get another pair of eyes on it before merging to gh-pages.

jeryj:
@joe-watkins Looks like the issue was with switching to the _data/resources (great idea, btw). I've got it fixed. I added the tabindex back in on index.html sections, and changed it to a -1 value, as well as changed the resources and patterns page tabindex section values to -1. That is a much better solution, so thanks for mentioning it!

If you have a chance, can you take a look and merge if everything looks good?

joe-watkins:
@jeryj slick! tnx ;)

PR #409, opened 2016-04-06
Updated URL

AllThingsSmitty:
The link to the mobiforge article was broken. Also removed the “10 Easy
Things All Web Developers Should Do” post; that site is no longer up


davatron5000:
Thanks, Matt! Very helpful. Good eye.

Issue #410, opened 2016-04-08
Add year for resource articles?


AllThingsSmitty:
Many of the articles under resources are 2-3 years old or older. Do you think there might be some benefit to adding the year to the title?

Ex:

- "Accessibility Checklist Article" by Cameron Moll (2008)

But this could also make the information feel dated. Thoughts?

davatron5000:
Maybe a separate issue, but I think it'd maybe be good to prune Resources from time to time as well. Dates would definitely help that. A11Y Project is less a "bucket of everything" and more "best of the best" content.

AllThingsSmitty:
Good point. I agree it's worth reviewing those further.

webuxr:
FYI: See also #319, #452, and #454.

In #454 I did attempt to not only maintain the year listed within some links but also attempted to come up with a consistent way of displaying it. For instance: 

- [Make Your Widgets Sing With ARIA](http://vimeo.com/73427435) - WDCNZ (2013)
- [Accessibility is My Favorite Part of the Platform](https://www.youtube.com/watch?v=2qjgxH384Nc) - Google I/O (2016)
- [PWAs in any context](https://www.youtube.com/watch?v=8dr_IUGwsO0) - Progressive Web App Summit (2016)

I can certainly see the need for a `event-date` or simply `year` token to help keep content separate from presentation and/or display characters (i.e.: the YAML file would contain "2016" verses "(2016)".

Which also leads me to think this resources.yaml file is turning into a nifty flat-file database or sorts.

**Thoughts about making links with a "posted on year" display consistently?**
**Additional thoughts about adding another token?**

Tagging people for visibility: @davatron5000, @jfc3-dol, @AllThingsSmitty, and @prettydiff. 

Cheers!

PR #416, opened 2016-04-18
Added year to resources

AllThingsSmitty:
This was mentioned in Issue #410, adding the year to resources when available to identify how old the material is. I think we’ll eventually get to reviewing all material for highest quality but for now this is a way way show what is recent.


davatron5000:
Thanks for being the Resource Czar! 🏆

PR #421, opened 2016-06-06
Improve markdown formatting on Color Contrast page

schpet:
Improve some issues with the markdown on the "What is Color Contrast?" including adding spaces so the Jekyll formatter renders lists and headings correctly.

http://a11yproject.com/posts/what-is-color-contrast


davatron5000:
Thanks! Not sure what happened with the titles there.

PR #433, opened 2016-08-29
headings with no space don't render correctly

paulcpederson:
screen shot 2016-08-29 at 10 18 45 am


davatron5000:
Thanks so much!

PR #435, opened 2016-09-01
fix heading markdown code

bransbury:
I've been reading through these - really great content by the way! - and noticed the heading weren't displaying correctly on this page. I think they just need a space after the ```##``` as this is what's on other pages. I've done the simple fix here.

For reference, here's what the page looks like for me before:

a11y-headings


davatron5000:
Thanks! These keep cropping up, if you find any more keep the PRs coming. 🙇

PR #439, opened 2016-10-13
Clean up old links

mkuehn:
Removed links that no longer exist, and updated redirect links.


davatron5000:
Thanks for reviewing those!

PR #442, opened 2016-10-22
updated link to modal window script

scottaohara:
hey there,

Saw that an older version of my modal window script was linked to on this page, so I updated it to the latest version.

Thanks!


scottaohara:
ah, reread the contribution docs and the new script isn't a codepen, so i'll close this for now.  sorry

Issue #429, opened 2016-08-04
Accessibility in print mode


sabarigr:
Hi,
A11Y Project is a good initiative to help developers like me to create accessible websites and applications. During my recent visit to my website I noticed that the urls in your website weren't made printer friendly, so i created this fiddle solving that.
Please visit https://jsfiddle.net/sabarigr/cqsp6n3f/

Thank you!
Kudos again to your initiative :+1:

jpdevries:
This is really clever! Speaking of print stylesheets, I'm curious if there is a general consensus on if using `!important` declarations for print style sheets is reasonable.

Example:
```css
@media print {
  * {
    background: white !important;
    color: black !important;
    border-color: grey !important;
  }
  .ads {
    display: none !important;
  }
}
```

sabarigr:
@jpdevries Yes. There is a specific usecase. If an element was styled via Javascript (_inline styles_), the same style would be applied to _print_ media as well. This is because browsers don't understand that the styles you declared through style attribute are for _screen_ media only.

**example:**
`<div style="opacity: 0;">An invisible division</div>
@media print {
  div {
    opacity: 1 !important;
  }
}`

Thanks :-)

jpdevries:
Thanks @sabarigr. Even if there weren't inline styling, if you wanted to use lose selectors in the print media query you would still need `!important`, no?

 I wrote up on some of my thoughts on practical, and maybe less practical, use cases for `!important` here https://medium.com/markuptips/css-is-important-960a9921e454#.mijz4hbgn

I like the High Contrast mode and Reduced Animation use cases

davatron5000:
Happy to support print mode for pages. Does anyone want to make a first pass at a Pull Request?

PR #444, opened 2016-10-30
Add Basic Print Stylesheet

mtthwhggns:
As per the discussion (https://github.com/a11yproject/a11yproject.com/issues/429), I've attempted a simple print stylesheet.

It's not meant to cover everything, but I've tried to make it look ok when an article is printed.

I've not done this before, so I'd appreciate as much feedback as possible!

Here's an example of a printed page with these styles applied; 
[MYTH_ Screen readers don’t use JavaScript - The A11Y Project.pdf](https://github.com/a11yproject/a11yproject.com/files/560374/MYTH_.Screen.readers.don.t.use.JavaScript.-.The.A11Y.Project.pdf)

### DIFF
--- a/_sass/_print.scss
+++ b/_sass/_print.scss
@@ -1,0 +1,45 @@
+@media print {
+  @page {
+    margin: 0.5cm;
+  }
+
+  article {
+    a:after {
+      content: ' ( ' attr(href) ' ) ';
+    }
+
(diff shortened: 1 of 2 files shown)


davatron5000:
Wow. This looks incredible! Thanks so much for taking the time to do this.

Issue #361, opened 2015-10-01
Alphabetizing links on resources page


AllThingsSmitty:
As the list of resources grows would there be any advantage/benefit to ordering the resources in each list alphabetically?

joe-watkins:
I'd be into that :+1:

AllThingsSmitty:
If we can get a consensus that would be nice but I may submit a PR for this, too.

davatron5000:
Totally down. Thanks, Matt.

AllThingsSmitty:
Submitted PR https://github.com/a11yproject/a11yproject.com/pull/362. Should we document this so contributors will know to follow the alphabetical pattern?

davatron5000:
Merged #362. Might be good to document our decision in `Contributing.md`

AllThingsSmitty:
@davatron5000 I'll work on adding a blurb about it there.

AllThingsSmitty:
Submitted PR https://github.com/a11yproject/a11yproject.com/pull/363 with this information.

PR #352, opened 2015-07-29
Update resources.html

indrekpaas:
Added "ARIA attribute and role completions for Sublime Text" link.

PR #445, opened 2016-11-02
Create resources page from a yaml data file

moarsel:
Related to #352 and #361, this PR moves the resources to a YAML file which is pulled into to the resource page and automatically alphabetically ordered.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -36,8 +36,7 @@ Here are some suggestions and tips on writing your article:
 
 Find a great resource that you think will be useful to others?
 
-- Add it to the [resources](https://github.com/a11yproject/a11yproject.com/blob/gh-pages/resources.html) page under the appropriate category.
-- Resources should be listed alphabetically.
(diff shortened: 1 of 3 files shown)


joe-watkins:
@moarsel This is cool -- it may be helpful for some contributors to know they can't use tabs in yaml. Perhaps a note about that in the contributing doc or a link to yaml's site? The file format is a bit less common than HTML and could freak out a new contributor :)

moarsel:
That makes sense @joe-watkins I'll see if I can document it a bit more clearly.

joe-watkins:
@moarsel I could see other sections of the site adopting this approach so maybe a general Yaml section in the contributing doc would work?

moarsel:
Ok @joe-watkins I took a first shot at it, let me know what you think.

moarsel:
Hey @joe-watkins, anything else needed to merge this in? If it makes sense, I'm also wondering if I should use this approach with the "patterns" section.

joe-watkins:
@moarsel Lookin' good! Tnx again. I think this same approach could work in Patterns as well... @davatron5000 what ya think?

davatron5000:
❤️ it! Let's merge it in.

PR #451, opened 2017-01-21
Grammar fixes.

willbiddy:


davatron5000:
Thank you so much, @willbiddy! We need all the help we can get with grammar fixes.

PR #458, opened 2017-01-28
Fixing broken Yahoo link.

willbiddy:
At some point Yahoo moved their blog to tumblr.


davatron5000:
Thanks so much @willbiddy

PR #462, opened 2017-02-01
Add "I want to see like the colour blind"

nschonni:
Updated version of #431

PR #431, opened 2016-08-09
Add "I want to see like the colour blind" Chrome extension to Resources

scriptype:
I find this extension to be very useful and simple.

BTW I highly appreciate this project and I'd be glad if I can do anything further to help.

PR #464, opened 2017-02-02
Add colour blindness Chrome extension to Resources

scriptype:
Hi, my previous PR was closed because of the technical issues mentioned there (#431). So, here's the new PR.

ps. By the time of writing this I wasn't aware of another PR #462 is opened to add this link. No problem if that PR is accepted instead.


joe-watkins:
Great addition @scriptype 👍

PR #465, opened 2017-02-04
docs(resources.yml): add Inclusive Design Patterns

felixzapata:
- Add new book: "Inclusive Design Patterns"


joe-watkins:
Good one!

PR #479, opened 2017-03-05
Improve code readability

swashcap:
#### Problem

Readability of of inline and block code on <a11yproject.com> isn’t great primarily because it’s set in the body type, a san-serif face. This is especially noticeable on [“Getting started with ARIA”](http://a11yproject.com/posts/getting-started-aria/). Code is traditionally set in a monospaced face.

#### Solution

* Address the issues by changing `code` elements’ `font-family` to a cross-OS-friendly monospace font stack.
* Decrease `code`’s `font-size` to make reading inline elements less jarring.
* Decrease `pre`’s `line-height` to make reading code blocks easier.
* _Bonus:_ remove redundant background color on `code`s nested inside `pre`s.

These are broken into (excessively?) granular commits. I’d be happy to rebase, change the font stack, drop a commit, etc.

#### Testing

```shell
git clone git@github.com:swashcap/a11yproject.com.git wat-is-code
cd wat-is-code
bundle install
jekyll serve
```

Then, go to http://localhost4000:/posts/getting-started-aria/ and check out the inline and block code.

#### Screenshots

Before:

before

After:

after

### DIFF
--- a/_sass/_typography.scss
+++ b/_sass/_typography.scss
@@ -88,6 +88,7 @@ pre {
 
 code {
     background: $grayLighter;
+    font-family: 'Source Code Pro', Consolas, Monaco, 'Courier New', monospace;
     padding: 0.1em 0.25em;
 }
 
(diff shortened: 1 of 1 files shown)


joe-watkins:
@swashcap great PR tnx :)

PR #480, opened 2017-03-07
Fix active logo bug

joe-watkins:
Removing active state on logo - was causing the logo to render incorrectly. :)


davatron5000:
@joe-watkins Just want to hop in and say GREAT JOB!! And thanks for all the work recently. 🎉

joe-watkins:
@davatron5000 thanks much :) that means a ton! This project is special to me and I owe it. Thank YOU for creating it!

PR #490, opened 2017-06-06
Add accessible error messages pattern

indrekpaas:
Adding @hiddedevries’ tutorial for creating accessible error messages. This PR does not follow the contribution guidelines by not including a live demo, but I’ve noticed a few other submissions have been accepted as blog posts/tutorials.

### DIFF
--- a/_data/patterns.yml
+++ b/_data/patterns.yml
@@ -91,6 +91,12 @@
     - url: "http://github.com/paypal/accessible-html5-video-player/"
       title: "Accessible HTML5 Video Player"
 
+- pattern-category: Error Messages
+  section-id: error-messages
+  patterns:
+    - url: "https://hiddedevries.nl/en/blog/2017-04-04-how-to-make-error-messages-accessible"
+      title: "How to make error messages accessible"
+
 - pattern-category: Other
   section-id: other
   patterns:


svinkle:
I'd prefer to have a working example before this submission is accepted. 

What's described in the article, using in-line validation, tabbing through and using an `aria-live="assertive "` region along with `aria-atomic` might result in a flurry of messages which could prove to be overwhelming for some.

I'm curious about the experience using this approach. @indrekpaas, if you're able to put something together I'll be happy to give it a review.

timwright12:
Agreed,`aria-live="assertive` might be a bit too strong. In most cases you can get away with `polite` (or nothing at all for static alerts) and not pulling the user away from what they're doing.

Found this about combining `role="alert"` with `assertive`
"Some people recommend adding a redundant aria-live="assertive" when using this role. However, adding both aria-live and role=alert causes double speaking issues in VoiceOver on iOS.

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions

hidde:
Hello! The author of the original post here. 

In a project I am working on we have actually tested this with an assertive live region, and it has worked well. It is a bit strong, but then so is our error messaging for sighted users (it is bold, red and quite visible). This may not fit every project and a conscious choice one could make.

A ‘flurry of messages’ could happen, depending on how the technique is implemented (i.e. if many errors are expected to happen, grouping them may be the more sensible approach).

In my defense, using assertive live regions is recommended in various places. [WebAIM](http://webaim.org/techniques/aria/) says:

>An aria-live value of assertive will result in the user being alerted to the content change immediately or as soon as possible. Assertive would be used for important updates, such as error messages

It is also recommended in the W3C's “Techniques for WCAG 2.0” documentation, as a [recommend technique to meet success criterion 3.3.1](https://www.w3.org/TR/WCAG20-TECHS/ARIA19.html).

Please let me know if this changes anything; I'd be happy to then add some live examples of the technique.

svinkle:
@hidde That would be awesome if you could provide a demo of your solution. I'm quite interested to experience what this type of workflow is like. It would be a great add to the patterns page as a working demo.

svinkle:
@timwright12 I read the same thing somewhere else. I did some testing and sure enough, simply adding `role="alert"` to an element announced the content in all major browser/screen reader combos. Great share. 👍

hidde:
@svinkle the article has now been updated with two demoes on Codepen to try out the approach (with different live region settings) https://hiddedevries.nl/en/blog/2017-04-04-how-to-make-error-messages-accessible

svinkle:
@hidde Thanks for making these demos! It helps to illustrate your point in the article. 🙂

## Combining attributes
Your note on _combining_ the attributes, I think this line in the article was where my confusion came from:

> There are two ways that are roughly equivalent, both making use of ARIA attributes: live regions (aria-live) and role=alert. 

It might be better to change to: "(aria-live) **or** role=alert"

## Demo 1
The first CodePen demo has one issue; when using `aria-*` you must provide a value. In this case just change this to `aria-invalid="true"`. Otherwise this results in invalid HTML. Using attribute selectors like this is pretty slick. 👍

## Demo 2
The second demo works well to get your point across. My concern with the described technique is whether this actually helps or hinders the user's ability to fill out the form, to understand how to fix the errors, and which inputs require fixing.

### aria-live="assertive" and role="alert"
The first two fields with `aria-live="assertive"` and `role="alert"` attributes, respectively, are similar in how the error message is announced directly after moving away from their related `input`. When doing so, the `label` announcement for the currently selected field is interrupted. For some users with cognitive disabilities this could be confusing as it may seem that the current field is in an error state. In order to hear the current field `label`, the user must move forward, or backward then forward. This could be quite cumbersome or an irritant for some people.

### aria-live="polite"
The last field with the `aria-live="polite"` attribute is better as it announces the error message _after_ the currently selected form control's accessible name is announced. Even still, this does require some extra effort on the user's part to move backward in the form to correct the issue.

### Error placement
The other thing worth mentioning is the placement of the error text; outside of the associated field `label`. This would make it difficult for users to understand the state of each field when navigating the form. 

I'd suggest placing the error text within the `label` to ensure it is announced on `input` focus. Also, in order to keep the current visual style, wrap the `input` and error text within the `label`. For example:

```html
<div class="question">
    <label for="question-1">What is your name? <em>Note, when you leave the field an error will appear in an explicit live region (assertive)</em>
        <input type="text" id="question-1" data-error="That's not a name" data-has-error="true">
        <div aria-live="assertive" aria-relevant="additions removals" class="error-area">
            <!-- insert error messages here -->
            <p>Error: That's not a name</p></div>
        </div>
    </label>
</div>
```

With the implicit `label` in place, the error text will be announced along side the `label` text. _Bonus_: this creates a much larger touch/click target which benefits people who may have difficulty with small, precise touch/click areas.

---

Without a usability testing session with real people and honest feedback, I'd be reserved to add this  to the patterns page. I'm currently in a position where I could set this up at some point in the future to get this feedback. I'll see about adding these examples in as part of an upcoming session.

In the meantime, let me know your thoughts on my feedback. See what you think about making the changes described above.

hidde:
Hi @svinkle! Thanks for your comments. 

Have updated the demos: 

* `aria-invalid` demo now has a value (nice spot!)
* live regions demo has more notes and have fields associated with their error area

They are still a simplification of what a real application would do (and what ours does), but at least they are now closer to what they should be and what works for users. 

I've also added a note to the introduction of the two attributes, to clarify they are best not used at the same time. Thanks for pointing out this was not clear to you.

My article was mostly to describe the idea of accessible error messaging, not so much to provide implementation details of The Single Best Way. On purpose, because what works best for a given implementation depends on lots of stuff, it is hard to give one perfect example without knowing a particular application's constraints. 

The above is why I had not originally included demos in the article and mostly summarised the concepts that we, in our tests, found worked well (and which, as I said before, are recommended by WebAIM and W3C). Of course, people can also use `polite` regions instead, or even use a different mechanism. 

Would be interested to hear if your user tests differ from ours! 

In the meantime, feel free to link or not link my article.

svinkle:
Just coming back to this. Unfortunately I'm not able to setup any user testing anytime soon. That said, I'd still like to have these CodePen demos as options for people to check out on the Patterns page.

👉 @indrekpaas, would you mind creating a new PR for the Patterns page, linking directly to @hidde's CodePen demos?

- [Styling `aria-invalid`](https://codepen.io/hidde/details/GEMaQz/)
- [Errors in live regions](https://codepen.io/hidde/details/eRGwer/)

👉 @hidde, I'd recommend linking to your article from the CodePen demos, either in the description or somewhere in the main body. This way people will have a chance to read more about the demos and why they exists.

Thanks! 🙂

PR #491, opened 2017-06-07
Formatting, additional info on scope

alterlife:
1. Cleaned up header indentation
2. Added additional info about scope=row and scope=col.
3. Fixed a couple of typos

note: I believe there's also 'colgroup' and 'rowgroup' for complex tables, perhaps that deserves a separate article?

### DIFF
--- a/_posts/2016-03-05-accessible-data-tables.md
+++ b/_posts/2016-03-05-accessible-data-tables.md
@@ -11,7 +11,7 @@ categories:
 
 The semantic purpose of a data table is to present tabular data. Sighted users can quickly scan the table but a screen reader goes through line by line. Proper markup must be added to help the screen reader make the correct associations that a sighted user would.
 
-# Example of an accessible data table.
+## Example of an accessible data table.
 {% highlight html %}
 	<table>
 	<caption>Monthly Budget</caption>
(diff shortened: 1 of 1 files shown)


_posts/2016-03-05-accessible-data-tables.md
```diff
@@ -42,24 +42,27 @@ The semantic purpose of a data table is to present tabular data. Sighted users c
...
+## Row and Column Headings
+For sighted users, row and column headings of tables are often highlighted visually. The screen reader will need this information coded into the markup.
 
-The `scope="row"` is typically put on the first `<th>` of the row. In this case <strong>"January" and "February"</strong> would get them. 
+Column headers should be marked using `scope="col"`. The `scope="col"` typically goes on the `<th>` element. If the `<th>` is a column like the <strong>"Amount Earned", "Amount Spent" and "Amount Saved"</strong> in my example, you would put them all as `scope="col"`. This lets the screen reader know that the `<th>`’s are column headers for that entire column. Column headers should be unique.
+
+Row headers should be marked up using `scope="row"`. The `scope="row"` is typically put on the first `<tr>` of the row. In this case <strong>"January" and "February"</strong> would get them. For optimal usability, the row header selected should be sufficiently unique and readable to identify the row.
```
svinkle: >  the first `<tr>` of the row
should be:
>  the first `<th>` of the row

alterlife: I think it should be `<td>` rather than `<th>` for row headers.

Good catch.

svinkle: Cell header with `scope="row"` would still be a `<th>`.

alterlife: https://www.w3schools.com/tags/att_td_scope.asp

alterlife: That's interesting, The <td> scope attribute is not supported in HTML5. Does that mean only the th is marked with scope and all the rows are automatically read as row headings?

alterlife: I did some more reading, It seems that my original understanding that th should not go into tbody is incorrect, it is recomended and valid for accessibility:
https://stackoverflow.com/questions/5509163/does-html5-ban-th-cells-from-tbody

I see that your example also covers this. Will you be editing the article?

svinkle: Update your PR for now with this change. Going to get more eyes on it before we accept.

alterlife: I've done so.

svinkle:
These changes look good to me.

/cc @Far3 @grayghostvisuals @joe-watkins

grayghostvisuals:
@svinkle Unfortunately I'm not able to donate time to the project right now. Thanks for including me though.

PR #506, opened 2017-08-15
Update 2013-01-14-alt-text.md

MarcusJT:
- Revised content to describe the common situation where an image and text are used together inside a link.
- Reshuffled the guidance into a more logical order, including bringing the basic purpose of alt attributes right to the top (from bottom).
- I also suggest the date should be removed from the filename but I'm not sure how to do that.


svinkle:
@MarcusJT Thanks for the update!

As for the date in the file name, Jekyll uses this meta data as a way to logically order the articles when the site gets built.

Issue #514, opened 2017-08-24
Revising TOC navigation markup


scottaohara:
there are two issues with the table of contents navigation items on pages like the index, resources and patterns:

1. marked up as `<section>`s rather than `<nav>` elements.  
2. these navigations contain the second heading on the page, but skip over the `<h2>` for the `<h3>`

I will be submitting a PR to address these issues

PR #515, opened 2017-08-24
revise toc nav markup and labeling

scottaohara:
[per issue #514](https://github.com/a11yproject/a11yproject.com/issues/514)  

• set table of contents navigations to be nav landmarks with an aria-labelledby pointing to their containted heading element.

• update heading helement from `<h3>` to `<h2>` so a heading level is not skipped in the document.


svinkle:
Good one. 👍 🛳

Issue #519, opened 2017-08-26
Skipped heading level on archive page


scottaohara:
The headings skip from `h2` to `h4` for the article listings on the [archive page](http://a11yproject.com/archives.html).

Will submit a PR to remedy this.

PR #520, opened 2017-08-26
fix heading order

scottaohara:
heading level 3 is currently skipped on the archive page.

this commit updates the CSS to include some utility classes, to allow us to modify font-sizes to match headings, without the need to use those headings.

It also fixes the heading element in the markup.

Per [issue 519](https://github.com/a11yproject/a11yproject.com/issues/519)

### DIFF
--- a/_sass/_typography.scss
+++ b/_sass/_typography.scss
@@ -37,11 +37,13 @@ h3 {
     font-size: 1.66em;
 }
 
-h4 {
+h4,
+.txt-up-2 {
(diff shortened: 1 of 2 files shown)


jeryj:
@scottaohara Thanks for catching and fixing the skipped heading order! For the classes, I'm fine with utility classes, but can we leave the "media-heading" class on the heading? Could be handy for future changes.

scottaohara:
@jeryj good call.  Added it back in. Thanks!

Issue #455, opened 2017-01-27
Checklist missing TOC side menu


webuxr:
When navigating between pages, I noticed the Resources page had a side TOC menu while the Checklist page did not ([see screenshot](https://screencast.com/t/dZeXdGSoNr)). Is this intentional or does the Checklist TOC just need to be built?

davatron5000:
@jeryj might know more about why this is triggers kinda late. I've seen it too from time to time.

jeryj:
@webuxr It was intentional because I felt like it was a giant form that was meant to be taken all in at once, rather than jumping from section to section via a sidebar. However, looking at it again, I can definitely see the rationale for having a sidebar. I'm open to either route (leaving it sidebar-less or adding one in). Do you have a strong preference about it either way?

webuxr:
I can see @jeryj 's pov and it makes perfect sense. My main concern is the visual jump experienced when navigating between the homepage and internal pages with a sidebar. For instance, going from Resources to Checklist or viewing the About page then going to the Patterns page.

The site has a homepage plus 4 high-level pages; 2 have the righthand sidebar nav (Patterns and Resources) while the other 2 (Checklist and About) plus the homepage, do not.

There are plenty of site's with multiple layouts (sidenav, no sidenav, 3-cols, etc.); that's not the real issue. And after looking a bit deeper into the issue I was seeing, it turns out not an issue with the sidenav at all. And so, I'm thinking about closing this issue and opening another explaining the "real" issue: the HTML/CSS grid.

The site's content region has a width of 828px.
When a page has a sidenav, that 828 is divided into a 625px content area and a 184px sidenav (with 19px gutter).
When a page _does not have a sidenav_, it's width is still the same 625px width, **but it's centered**. This is what is causing the visual jump (not having a sidenav or not).

In other words, #455 is NOT a bug, it was built that way on purpose.

So, I'll leave it up to @davatron5000 as to whether I should close this issue (and open a new issue to address the grid alignment) or leave this issue open (and re-word it's title and labels). Thoughts?

webuxr:
Slight correction...
The homepage uses a totally different layout with a lefthand sidenav area. It's essentially a flipped version of the Resources layout but with a smaller (width) sidenav. This layout divides into the following: a 147px sidenav area plus the same 625px content area. This variation does not help the visual jump issue originally noted.

joe-watkins:
@webuxr Sounds like we could do these things:

[...]

PR #524, opened 2017-09-03
Address layout changes in #455

joe-watkins:
Addresses layout changes in #455

- [x] Ensure all 2 column layouts the same.
- [x] Make home page 2 column layout the same as inner pages.
- [x] Move home page TOC to the right (for consistency).
- [x] Add a 2 column layout with TOC to Checklist page.

I've also removed the font sizing on active right column TOC elements as it was making text jump on longer items to two lines. I've instead added text underline for visual active state.

### DIFF
--- a/_layouts/archives.html
+++ b/_layouts/archives.html
@@ -25,5 +25,7 @@ layout: default
 </div>
 
 <article class="post">
-  {{ content }}
+  <div class="post-content">
+    {{ content }}
+  </div>
 </article>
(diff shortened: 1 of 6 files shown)


svinkle:
We might want to have a label the `fieldset`s for when they receive focus. Something like:

```html
<fieldset class="article-section resources-section" id="aria-roles" aria-labelledby="landmarks-section" tabindex="-1">
    <legend id="landmarks-section">Landmarks</legend>
    <!-- … -->
</fieldset>
```

This can be a different PR though.

joe-watkins:
@svinkle tnx for review and great idea about the label for the `fieldset` I'll fire up a PR for that.

Issue #12, opened 2013-01-13
List of Post Ideas


davatron5000:
Simple, bite-sized snippets of content to teach A11Y in smaller pieces instead of big chunks. We can check them off (using githubs new todo list feature) as we get them finished.

:chart_with_upwards_trend: == A table or  chart here would be useful.
## Concepts

Basic introductions to various concepts. What is it? Why does it exist?
- [x] ARIA introduction: roles :chart_with_upwards_trend: 
- [x] ARIA introduction: states :chart_with_upwards_trend: 
- [x] ARIA introduction: properties :chart_with_upwards_trend: 
- [x] Basic ARIA roles
- [x] Basic ARIA states
- [x] Basic ARIA properties
- [x] Color blindness
- [x] Color Contrast
- [ ] What is WAI? The Web Accessibility Initiative
- [x] Section 508
- [x] WCAG 2 (38 reqs)
## Features
- [x] Accessibility Checklist
- [x] Resource Page
## Quick Tips

Quick, tiny, easy improvements for maximum ROI
- [x] Using `alt` text correctly
- [x] Using `title` attributes correctly
- [x] Never use `maximum-scale=1.0`
- [x] Prefer button to div elements [https://staging.w15y.com/shortcuts/prefer-button-to-div](shhhh!)
- [ ] Avoid -webkit-text-adjust-size: none; http://codepen.io/davatron5000/full/14/7
## Quick Tests
- [x] Navigate using your keyboard
- [x] Test contrast on your phone
- [x] Larger Touch Targets
- [x] Accessible Typography (considerations: larger font-size, etc)
## HOWTOs

More in-depth, yet simple articles on HOW-TO use/create something.
- [x] HOWTO: Use VoiceOver on iOS
- [x] HOWTO: Use NVDA on Windows
- [ ] HOWTO: Use JAWS on Windows
- [x] HOWTO: Use VoiceOver on Mac
- [ ] HOWTO: Make an accessible `<nav>` with dropdowns @jimbateson
- [x] HOWTO: Hide content
- [x] HOWTO: Write accessible `FORM`s
- [x] HOWTO: Make accessible `VIDEO`s: `<track>` & HTML5 Video Controls
- [ ] ?? HOWTO: Make accessible `CANVAS` & SVG ?? (2 posts?) @CBID2 
- [ ] HOWTO: Ajax with `ARIA-LIVE`
- [ ] HOWTO: Make accessible dialogs and modals
[...]

PR #532, opened 2017-09-14
Remove role="navigation" from example

albertoconnor:
There is some contention about whether or not `<nav>` is semantically the same as `<nav role="navigation">`, but the October 2014 version of the HTML spec seems to indicate it shouldn't be set:

https://www.w3.org/TR/html5/sections.html#the-nav-element

Even if `<nav>` implicitly having the "navigation" landmark isn't fully implemented by all screen readers, I would think this documentation should be aspirational, showing how things will be.

Having `<nav role="navigation">` contradicts the important, and I think generally hard to grasp the idea, that redundant roles should not be used. Part of the reason it is hard to grasp due to a proliferation of incorrect examples.

That being said having `role="navigation"` still may be practical and a good idea currently, but if so my feeling is that should be explained.

### DIFF
--- a/_posts/2014-05-15-getting-started-aria.md
+++ b/_posts/2014-05-15-getting-started-aria.md
@@ -64,7 +64,7 @@ To create accessible applications, basic principles of semantic HTML, keyboard s
 ## ARIA Examples
 
 * Landmark role
-  The `<nav>` element has been given a landmark role allowing screen reader users navigate directly to this element.
+  The `<nav>` element implicitly has landmark role of `navigation` allowing screen reader users navigate directly to this element. [Discussion on implicit html5 landmarks](http://a11yproject.com/posts/aria-landmark-roles/)
 
 ~~~~~~~~
     <nav>
--- a/_posts/2014-05-15-getting-started-aria.md
+++ b/_posts/2014-05-15-getting-started-aria.md
@@ -64,7 +64,7 @@ To create accessible applications, basic principles of semantic HTML, keyboard s
 ## ARIA Examples
 
 * Landmark role
-  The `<nav>` element implicitly has landmark role of `navigation` allowing screen reader users navigate directly to this element. [Discussion on implicit html5 landmarks](http://a11yproject.com/posts/aria-landmark-roles/)
+  The `<nav>` element implicitly has a landmark role of `navigation` allowing screen reader users to navigate directly to this element. Review the article [Quick Tip: Aria Landmark Roles and HTML5 Implicit Mapping](http://a11yproject.com/posts/aria-landmark-roles/) for more information. 
 
 ~~~~~~~~
     <nav>
--- a/_posts/2014-05-15-getting-started-aria.md
+++ b/_posts/2014-05-15-getting-started-aria.md
@@ -67,7 +67,7 @@ To create accessible applications, basic principles of semantic HTML, keyboard s
   The `<nav>` element has been given a landmark role allowing screen reader users navigate directly to this element.
 
 ~~~~~~~~
-    <nav role="navigation">
+    <nav>
       <ul>
         <li>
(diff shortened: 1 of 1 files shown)


scottaohara:
Hi @albertoconnor,

I agree that in most cases the `role="navigation"` doesn't need to be set, but for the reasons that the role of `nav` is well supported in modern browser + AT combos.  It's more of unnecessary redundancy than anything at this point.

Per your PR, you removed the `role="navigation"` from the code example, but the paragraph above the example still references the role being added.  Would you want to take a stab at rewriting that bit?

Thanks!

albertoconnor:
Thanks for the feedback @scottaohara, I have taken a stab at it, but I am not sure the best way to link internally.

I discovered a more detail discussion on implicit versus explicit landmark roles on the site, I think linking to it would be handy if the example changes or even if it is decided it shouldn't be.

Thanks,
Albert

scottaohara:
Thanks @albertoconnor 

@svinkle would you like to take another look for any additional feedback?

_posts/2014-05-15-getting-started-aria.md
```diff
@@ -64,10 +64,10 @@ To create accessible applications, basic principles of semantic HTML, keyboard s
 ## ARIA Examples
 
 * Landmark role
-  The `<nav>` element has been given a landmark role allowing screen reader users navigate directly to this element.
+  The `<nav>` element implicitly has landmark role of `navigation` allowing screen reader users navigate directly to this element. [Discussion on implicit html5 landmarks](http://a11yproject.com/posts/aria-landmark-roles/)
```
svinkle: > implicitly has landmark role of `navigation` allowing screen reader users navigate directly to this element

👇

> implicitly has **a** landmark role of `navigation`**,** allowing screen reader users **to** navigate directly to this element

svinkle: > [Discussion on implicit html5 landmarks](http://a11yproject.com/posts/aria-landmark-roles/)

👇

> Review the article [Quick Tip: Aria Landmark Roles and HTML5 Implicit Mapping](http://a11yproject.com/posts/aria-landmark-roles/) for more information.

svinkle:
@scottaohara I think this raises a larger question of whether we should update all other content on the site where it suggested to explicitly add `role` attributes for landmarks. At the time it was required for older browser+AT support, but these days things are _almost_ fully supported (VO + `footer` comes to mind…)

scottaohara:
@svinkle yeh that's a good point. 

So we could either add notes to these articles that mention these roles are redundant, but useful if you know you have a group of visitors that are on old IE and possibly older ATs, the latter obviously being harder to 'know'.

Or yeh, we could divy up some of these articles and do minor rewrites.

i'm fine with either

albertoconnor:
I am happy to do a bit more of the editing, on this page if you want to go ahead with changes across the site, just let me know and I will apply the suggested changes and anything else you need to be done on this page.

One of the other reasons I brought this up was the page itself emphasizes that applying landmark roles to semantic elements is redundant:

> Document roles provide a structural description for a section and are typically non-interactive. Some document roles map onto existing HTML tags (Eg. <div role="form">) and are only meant for cases when using the native tag is not possible. It's unnecessary to add ARIA roles to existing semantically meaningful elements. For example, <form role="form"> is redundant.

Which then goes and does in the example I pointed out.

svinkle:
@albertoconnor Fixup the issues I pointed out with this first so we can get this PR merged in, then we can review other content pieces.

albertoconnor:
I will make the changes on Monday of next week!

albertoconnor:
I believe I have made the two changes, how else can I help?

svinkle:
Thanks, @albertoconnor! Check out #12 for ideas or anything else you'd like to help with would be great!

PR #541, opened 2017-10-18
update 2017-10-18-understanding-colourblindness.md

mgnc2867:
Fixed small wording errors.
Added hyperlink example.

### DIFF
--- a/_posts/2017-10-18-understanding-colourblindness.md
+++ b/_posts/2017-10-18-understanding-colourblindness.md
@@ -12,13 +12,11 @@ Some peoples' eyes can't tell the difference between certain colours; that's not
 
 Yet, colour blindness is not rare. Some form of colour blindness affects at least one in twenty people; [Wikipedia has a helpful table](https://en.wikipedia.org/wiki/Color_blindness#Epidemiology), based on [a journal article](https://apfmj.biomedcentral.com/articles/10.1186/s12930-014-0010-3). There are [several different forms](http://www.colourblindawareness.org/colour-blindness/types-of-colour-blindness/). The most common, red-green colourblindness, affects around one in twelve men.
 
-Colour blindness is less profound than many impairments. Yet, people with colour vision problems often face restrictions on certain activities. This varies around the world. For example, Romania prohibits residents who are colour blind from driving. Yet, the USA doesn't consider colour blindness as a disability, for workplace discrimination purposes.
-
-I've avoided going into the different types here, as the principles are the same for each. Also, [Wikipedia's article on colour blindness](https://en.wikipedia.org/wiki/Color_blindness) has some excellent descriptions.
+Colour blindness is less debilitating than many impairments. Yet, people with colour vision problems often face restrictions on certain activities. This varies around the world. For example, Romania prohibits residents who are colour blind from driving. Yet, the USA doesn't consider colour blindness as a disability, for workplace discrimination purposes.rely've avoided going into the different types here, as the principles are the same for each. Also, [Wikipedia's article on colour blindness](https://en.wikipedia.org/wiki/Color_blindness) has some excellent descriptions.
 
 ## What does this mean for accessibility on the web?
 
-Don't use on colour alone to communicate. As a result, you don't risk making life difficult for colour blind users. 
+Don't rely on colour alone to communicate. As a result, you can make life simpler for colour blind users. An example of this would be hyperlinks being designed to be underlined by default.
 
 Sticking to this is also compliant with the WCAG; guideline [1.4.1 covers making content distinguishable using colour](https://www.w3.org/TR/UNDERSTANDING-WCAG20/visual-audio-contrast-without-color.html).
 


_posts/2017-10-18-understanding-colourblindness.md
```diff
@@ -12,13 +12,11 @@ Some peoples' eyes can't tell the difference between certain colours; that's not
...
 
-Colour blindness is less profound than many impairments. Yet, people with colour vision problems often face restrictions on certain activities. This varies around the world. For example, Romania prohibits residents who are colour blind from driving. Yet, the USA doesn't consider colour blindness as a disability, for workplace discrimination purposes.
-
-I've avoided going into the different types here, as the principles are the same for each. Also, [Wikipedia's article on colour blindness](https://en.wikipedia.org/wiki/Color_blindness) has some excellent descriptions.
+Colour blindness is less debilitating than many impairments. Yet, people with colour vision problems often face restrictions on certain activities. This varies around the world. For example, Romania prohibits residents who are colour blind from driving. Yet, the USA doesn't consider colour blindness as a disability, for workplace discrimination purposes.rely've avoided going into the different types here, as the principles are the same for each. Also, [Wikipedia's article on colour blindness](https://en.wikipedia.org/wiki/Color_blindness) has some excellent descriptions.
```
mtthwhggns: Is this one a mistake?

_posts/2017-10-18-understanding-colourblindness.md
```diff
@@ -12,13 +12,11 @@ Some peoples' eyes can't tell the difference between certain colours; that's not
...
 
 ## What does this mean for accessibility on the web?
 
-Don't use on colour alone to communicate. As a result, you don't risk making life difficult for colour blind users. 
+Don't rely on colour alone to communicate. As a result, you can make life simpler for colour blind users. An example of this would be hyperlinks being designed to be underlined by default.
```
mtthwhggns: Thanks for the first one, that should definitely be fixed. 

Regarding the second one (the example), This isn't about making life simple. It's about not making tasks difficult (or impossible), and I don't think they are opposites.

svinkle:
@mtthwhggns I'll wait for your 👍 before merge.

_posts/2017-10-18-understanding-colourblindness.md
mgnc2867: Yup!

_posts/2017-10-18-understanding-colourblindness.md
mgnc2867: no problem :) I get where you're coming from. As a colour blind (not red-green) person, I've definitely had experience of stuff being made unnecessarily difficult.

mtthwhggns:
@mgnc2867 I've split this up now, so I've made a separate PR to fix just the wrong word (thanks again for noticing that one!), as I suspect we are in different time zones and didn't want to rush you!

I'd be happy to include the hyperlink example. Could you amend your commit to include only this?

I'd suggest a more active, simpler sentence, so something like;

> For example, hyperlinks are underlined by default, as well as a different colour. 

Let me know what you think.

mgnc2867:
I'll get on it now, I definitely think we're in different time zones!

mgnc2867:
Ready for your perusal.

_posts/2017-10-18-understanding-colourblindness.md
```diff
@@ -18,7 +18,7 @@ I've avoided going into the different types here, as the principles are the same
 
 ## What does this mean for accessibility on the web?
 
-Don't use on colour alone to communicate. As a result, you don't risk making life difficult for colour blind users. 
+Don't rely on colour alone to communicate. As a result, you don't risk making life difficult for colour blind users. For example, hyperlinks are underlined and coloured.
```
danielmatthew: This would read better as: 

> …you avoid making life difficult…

Something about `don't risk` feels awkward.

Not something you've changed in this PR, but perhaps while it's open?

_posts/2017-10-18-understanding-colourblindness.md
```diff
@@ -12,11 +12,13 @@ Some peoples' eyes can't tell the difference between certain colours; that's not
 
 Yet, colour blindness is not rare. Some form of colour blindness affects at least one in twenty people; [Wikipedia has a helpful table](https://en.wikipedia.org/wiki/Color_blindness#Epidemiology), based on [a journal article](https://apfmj.biomedcentral.com/articles/10.1186/s12930-014-0010-3). There are [several different forms](http://www.colourblindawareness.org/colour-blindness/types-of-colour-blindness/). The most common, red-green colourblindness, affects around one in twelve men.
```
joe-watkins: I recommend referencing the term "Color Vision Deficiency" somewhere as that is more of an accurate term somewhere in the post. https://www.aoa.org/patients-and-public/eye-and-vision-problems/glossary-of-eye-and-vision-conditions/color-deficiency

joe-watkins:
@mtthwhggns Made a comment about maybe adding "Colour Vision Deficiency" as a term in the post and can you check this sentence:

```
It might sounds understandable to show items that need attention in red, and those that are good in green. 
```
That read funky to you?

_posts/2017-10-18-understanding-colourblindness.md
mtthwhggns: I don't agree, I'm afraid. I spent a while thinking about this when drafting. 

I think `don't risk` feels most literal. Let me try and explain.

This comes down to the fact that there are many types of colourblindness, and most people with some form of colourblindness can tell the difference between many colours. 

If this were to read `you avoid making life difficult`, that gives the impression that if you get your colours wrong, it affects everyone with different colour vision. If this were the case, and one person with a colour vision impairment could see it correctly, it would suggest that there's actually nothing wrong with the colour choice.

On the other hand, if we say that you _risk_ making it difficult to understand, rather than implying you'll definitely make it difficult to understand, we make it clear that you can't be sure if you have an issue based on one person's experiences. 

tl;dr version: I was concerned that this suggestion sounded too binary, rather than reflecting the spectrum of different types of colour vision. 

If you have a suggestion for a way to make this clear in as short and simple a sentence, I'd be keen to hear it!

_posts/2017-10-18-understanding-colourblindness.md
mtthwhggns: I thought about this too, but was concerned it was too much like jargon. Most lay people know someone who is "colourblind", and most people use the term, so it seemed more reader-friendly to stick to the original.

mtthwhggns:
@joe-watkins I've fixed that in PR #543 , thanks!

_posts/2017-10-18-understanding-colourblindness.md
mtthwhggns: I've also addressed this in #543

mtthwhggns:
@mgnc2867 The example text seems fine, if you can rebase onto the more recent changes, this is good to go.

mgnc2867:
I'll sort this out this evening. Sorry for slowness!

svinkle:
@mgnc2867 @mtthwhggns Where are things with this PR? Still more work to be done?

_posts/2017-10-18-understanding-colourblindness.md
scottaohara: I actually agree with the awkwardness of "don't risk", but also agree with your reasoning for trying to convey the idea this way. 

I think instead of "don't risk" it should be "won't risk" as that is a more natural way to chain the sentences together. Cause and effect really.  Right now it reads as two commands, but the second is meant to be the outcome of the first "don't".  e.g. "don't do this so this won't happen."  Not "don't do this so this don't happen."

"Don't rely on colour alone to communicate. As a result, you *won't* risk making life difficult for colour blind users."

On a different note, "making life difficult" may be a bit hyperbolic, as it's really a difficulty in understanding their present intake of content. Their "life" (hopefully) hasn't gone down the tubes because the "username" label of a sign in field doesn't meet WCAG color contrast :)

mtthwhggns:
The only thing left is whether or not to add the hyperlink example.

It's a valid example, and can be added, although it's not essential. 

The only problem is that it needs a rebase, as the typo fix got merged first, so now we have a conflict.

svinkle:
@mtthwhggns I've fixed the conflict. What's the issue with adding the link? Adding links to further illustrate the point/argument is ideal.

PR #545, opened 2017-10-25
Events

ericwbailey:
This PR adds an events page to the primary navigation 🎉 

Right now I'm delineating between remote and physical events, so things like webinars versus events you need to be present for in person. Happy to update it if people are feeling the terms or division aren't needed. 

I've also added two events I know of off the top of my head, but we could definitely use more!

### DIFF
--- a/_data/events.yml
+++ b/_data/events.yml
@@ -1,0 +1,13 @@
+- event-category: Remote
+  section-id: remote
+  events:
+    - url: https://www.inclusivedesign24.org/
+      title: "Inclusive Design 24 (#ID24)"
+      time: 2017-11-16
+
+- event-category: Physical
+  section-id: physical
+  events:
+    - url: http://camp.a11yto.com/
+      title: "#a11yTOCamp"
+      time: 2017-11-18
--- a/_data/events.yml
+++ b/_data/events.yml
@@ -1,13 +1,25 @@
-- event-category: Remote
-  section-id: remote
+- event-category: Upcoming
+  section-id: upcoming
   events:
-    - url: https://www.inclusivedesign24.org/
-      title: "Inclusive Design 24 (#ID24)"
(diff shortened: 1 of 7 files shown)


svinkle:
@ericwbailey Awesome work!

I'm thinking the "Let us know" link should just point to the [new issue](https://github.com/a11yproject/a11yproject.com/issues/new) screen here on GitHub. If someone's not comfortable with making a PR, one of us can take the ticket and update the page for them.

I'd like to see more options available, in case anyone has more details they'd like to share. For example, check out the [CodePen Meetups](https://blog.codepen.io/meetups/) page. Maybe also include a data node for a description, an image/logo plus alt text, etc. We could get creative with the layout here too, not just a list but a data table if needs be.

Thoughts?

svinkle:
Also just thought, if an event is past the current date, should it just automatically be removed from the page? Or would there be a sub-page of archived events? I imaging an archive would be simple enough; just link from this page and have the same layout, but only include data with a date less than now type deal.

ericwbailey:
Good call on pointing to the issue page, I'll definitely update that. I also like including more event information and an archive of past events. 

For now, I'm going to just have a list at the bottom of the page for any event that occurred before the current day. We don't have many events right now, so it might help to see what was out there and also hold off on the complication of introducing another page until we need it.

As for event info, I'll see about ~stealing~ introducing some of the details the CodePen meetup page uses.

ericwbailey:
Okay, we now have a more fleshed-out event schema, and with a little JS help from a friend, an event archive  🎉

Since the site doesn't do a daily build, the JS organizes the list of events into future and past events based on the user's device date. If JS doesn't load, it is just one long chronologically-ordered list.

I'll probably tackle event images a little later, but I don't see that blocking a merge?

Issue #494, opened 2017-06-23
[Question][Suggestion] Never use maximum scale


kamlekar:
In this [article](https://github.com/a11yproject/a11yproject.com/blob/gh-pages/_posts/2013-01-14-never-use-maximum-scale.md), it is clearly said to avoid use of `maximum-scale` but what about `user-scalable`?

I think we should also mention about it in the same article so that a user won't get misdirected by doing one thing correct and other one wrong.

svinkle:
From [MDN: `<meta>`](https://developer.mozilla.org/en/docs/Web/HTML/Element/meta):

Value | Possible subvalues | Description
-- | -- | --
`user-scalable` | `yes` or `no` | If set to `no`, the user is not able to zoom in the webpage. The default is `yes`. Browser settings can ignore this rule, and iOS10+ ignores it by default.

By default, `user-scalable` is `yes` so the only reason someone would include it is to disable user zoom.

Is your suggestion to add this to the article to make sure someone _doesn't_ include `user-scalable="no"`?

svinkle:
@kamlekar, any comment on my question above?

PR #546, opened 2017-10-31
update maximum-scale article

dhcodes:
Per #494 updated article to discuss user-scalable attribute.


svinkle:
Thanks, @dhcodes! 👍

PR #556, opened 2017-11-17
Fixing small typo in placeholder article.

tjdunklee:
Thank you for this wonderful resource. I noticed the word "disappears" was spelled wrong in an article so I thought I would help out quick. Thanks again!


ericwbailey:
Ooh, great catch! Thank you for the fix 🎉

Issue #531, opened 2017-09-06
Create badge/pill design system


joe-watkins:
As noted in https://github.com/a11yproject/a11yproject.com/issues/513#issuecomment-324658307 we are needing a cool / easy-to-use badge/pill design system integrated into the site.

Will try this system out on the patterns page. Needs to be flexible so that it works in other areas of the site ;)

galantlex:
👋 Hi, I'd love to try to tackle this!

svinkle:
Hey @galantlex! Thanks so much, that would be awesome! Let me know your thoughts on what a solution might look like.

Looking forward to hearing what you come up with! 🙂

svinkle:
(I can't assign this ticket to you directly @galantlex but I'll assign to myself and we can work on this together.)

galantlex:
I'm pretty new to the world of design in general, so just to be clear, when you say "pill/badge", do you mean maybe something like this?

![a11y screenshot showing possible pill design](https://user-images.githubusercontent.com/17116584/31063247-587d496a-a6ff-11e7-8d38-a4f5258828be.png)

svinkle:
@galantlex Yeah, that looks pretty great to me!

I'm thinking it would be cool to just have the icons visible beside the link text, then on link `:hover`+`:focus`, display the framework name. This way the focus is more on the link itself and what it is, rather than the tech behind it.

galantlex:
Makes sense! In that case do you think there should be a border around it at all, or really just the icon? Especially since some icons (react, angular) seem more self-encapsulated than others (vanilla).

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

svinkle:
@galantlex Good point. I think just the icons only would look pretty slick. The vanilla JS icon could be the yellow square with black "JS" text, as seen on many a laptop. 🙂

kendrick:
@galantlex Happy to chip in on this if you need help/opinions!

jenstrickland:
Also, happy to help here. I wanted to ask if color contrast is a concern for the icons.

svinkle:
@jenstrickland Excellent! For contrast, we can't do anything when using brand logos. However, if we add plain text or a custom border for the design, that's when we need to account for color contrast.

@galantlex, curious if you were able to make any more headway with this? If not, no worries!

I'd like to see a decision made for a design concept first before anyone takes on the coding bit.

[...]

PR #567, opened 2017-12-21
Add 'code' field to patterns.yml, fill in some

galantlex:
Starting some work on the pill/badge design system, per discussion in #531

### DIFF
--- a/_data/patterns.yml
+++ b/_data/patterns.yml
@@ -3,113 +3,146 @@
   patterns:
     - url: "https://codepen.io/grayghostvisuals/pen/ltjDa"
       title: "Dropdown"
+      code: Javascript, CSS
     - url: "https://adobe-accessibility.github.io/Accessible-Mega-Menu/"
       title: "Mega Dropdown"
+      code: jQuery
     - url: "https://codepen.io/svinkle/pen/LDKwA"
       title: "Mobile Toggle"
+      code: Javascript, CSS
     - url: "https://codepen.io/grayghostvisuals/pen/IkmGF"
       title: "Off Canvas Push"
+      code: Javascript, CSS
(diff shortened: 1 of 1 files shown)


galantlex:
@svinkle, @jenstrickland, these are my very first steps: just adding the `code:` field to each pattern, and starting to fill in a few. For the ones that aren't associated with a framework, do we want to do `HTML, CSS, Javascript`, depending on whether they actually use all three? Or maybe make a third thing, that's just like `no framework`, but nicer-sounding? 😁

ericwbailey:
Looking really good so far, thank you!

"Vanilla" seems to be the go to term for frameworkless JavaScript, but it might be better to just stick with the underlying tech (HTML, CSS, JavaScript, etc.)? My thinking is if you don't know that Vanilla is an industry term, it could be confusion to beginners?

scottaohara:
to @ericwbailey's point:
Anecdotally speaking, I've run into various junior developers who were utterly confused when I mentioned "vanilla javascript".  wondering if i was talking about a new framework. 

just "JavaScript" or "ES#" seemed to be perfectly understandable though.

shrug

svinkle:
To me, "vanilla" = "JavaScript". Anything not vanilla would be called by its name; jquery, angular, etc.

That said, I also think "vanilla" also = ES6. I know some folks are still getting ramped up on ES6 features but pretty well all (modern) browsers have support so I don't think it's work differentiating.

@galantlex I think using what you've got here is great! It's what is what I was thinking as well. 👍

svinkle:
We probably shouldn't merge until #568 is in.

galantlex:
As a somewhat junior developer, yeah, I had no idea VanillaJS wasn't some other new framework 😁 

I'll continue on with HTML, CSS, JS, etc, and I'll make sure to test integration with #568 before merging in 👍

PR #568, opened 2017-12-21
add svg logos and a few png fallbacks

jenstrickland:
Add svg logos for CSS3, Angular, JavaScript, jQuery, NodeJS, React, and SASS, plus png fallbacks for CSS3, jQuery, NodeJS, and SASS


ericwbailey:
I'd like to see treatments that hold up at smaller sizes. Since the badging will be icon-sized, I worry that for more complicated logos such as jQuery and Node, a lot of the detail will be lost. Referencing each site's favicons might be a good place to start.

That being said, I do think it's good to have full-size versions of the logo available as well!

svinkle:
Agreed with @ericwbailey. 

I'll share a secret with y'all that I do. When I need an SVG logo, I head over to [IcoMoon](https://icomoon.io/app/#/select), do a search, and inspect the DOM and just grab the SVG code from there. They have a sweet selection available and since they're meant to be icon fonts, they'd all be appropriately sized for our needs.

@jenstrickland Try this out and see what you can find. 🙂

jenstrickland:
@ericwbailey @svinkle I love the favicon idea, and WOW! IcoMoon had fantastic icons for nearly all the options (Angular was missing so I grabbed their favicon). Much better sizes than I'd located. Thanks! Now that I know creating a separate scss file is the way to go I'll work on creating styles to pair with @galantlex 's work.

PR #572, opened 2018-01-09
Update 2013-01-11-myth-accessibility-is-blind-people.md

flooyd:
Changed language in driving analogy...for safety reasons :P

Also, I have never done a PR before so if I did something wrong let me know.


scottaohara:
looks good to me.  people definitely should not be looking at websites while driving :)

PR #575, opened 2018-01-15
Updates to the a11y checklist

scottaohara:
This commit has various changes to the checklist to modernize and correct some of the information.

* Revise ARIA landmark roles introduction paragraphs to point to more recent sources, and reword to indicate that older browsers may need landmark role support. (previously mentioned IE and iOS browsers specifically. Revised to just mention “older browser” to not confuse the fact that modern iOS browsers and IE11 have support for landmarks)

* Revise nav role description to be more descriptive.

* Update article description to note that it is not actually a landmark, but is treated similarly to one by some screen readers.

* Other spelling, capitalization, or sentence re-wordings that retain their original meaning.

### DIFF
--- a/checklist.html
+++ b/checklist.html
@@ -13,7 +13,7 @@ permalink: checklist.html
         <li><a class="outline" href="#outline">Document Outline</a></li>
         <li><a class="links" href="#links">Links</a></li>
         <li><a class="images" href="#images">Images</a></li>
-        <li><a class="js" href="#js">Javascript</a></li>
+        <li><a class="js" href="#js">JavaScript</a></li>
         <li><a class="forms" href="#forms">Forms</a></li>
         <li><a class="audio" href="#audio">Media (Audio and Video)</a></li>
         <li><a class="color-contrast" href="#color-contrast">Color and Contrast</a></li>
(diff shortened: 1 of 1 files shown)


jeryj:
Thanks for working on this @scottaohara!

checklist.html
```diff
@@ -214,17 +223,19 @@ <h2 id="toc_title" class="toc-title">Categories</h2>
...
+		<p class="description">Test against different types of color blindness with a tool like <a href="http://colorfilter.wickline.org/">http://colorfilter.wickline.org</a>. If you are on a Mac, another option is <a href="http://michelf.ca/projects/sim-daltonism/">Michel Fortin's, Sim Daltonism</a> color blindness simulator.</p>
   </fieldset>
 
 	<!-- Testing -->
 	<fieldset class="article-section resources-section" id="testing" aria-labelledby="label-testing" tabindex="-1">
 		<legend id="label-testing">Testing</legend>
-		<p>Navigating your site using a range of tools, such as just the keyboard or a screen reader, will help you understand how a blind, low-vision, or limited-mobility user will experience it.</p>
+		<p>
+      Navigating your site or application using a range of tools. For instance: using only a keyboard or a screen reader will help you understand how a blind, low-vision, or limited-mobility user will experience it.
```
svinkle: > **Navigate** your site or application using a range of tools.

> For instance**,** using

checklist.html
```diff
@@ -27,8 +27,15 @@ <h2 id="toc_title" class="toc-title">Categories</h2>
...
-		<p>Note: When you <a href="https://validator.w3.org/">validate html</a> using landmark roles, you'll receive a warning stating these roles are redundant.  In HTML5, several of the landmark roles are implicit via the native structural element which is supported by most modern <a href="http://stevefaulkner.github.io/html-mapping-tests/">desktop browsers</a> with the exception of IE and <a href="https://dequeuniversity.com/assets/html/jquery-summit/html5/slides/landmarks-example.html">iOS Safari</a>. So, if you support IE and iOS browsers, you'll want to use the landmark roles. For more information, read <a href="/posts/aria-landmark-roles/">Quick Tip: Aria Landmark Roles and HTML5 Implicit Mapping</a>.</p>
+		<p>
+      ARIA Landmark roles can help assistive technology users to quickly navigate to and past blocks of content in a web interface.
+    </p>
+		<p>
+      <strong>Note:</strong> When you <a href="https://validator.w3.org/">validate HTML</a> that utilizes most landmark roles, you'll receive a warning stating these roles are redundant.
```
svinkle: Should we update this to use the [Nu checker](https://validator.w3.org/nu/)?

checklist.html
```diff
@@ -27,8 +27,15 @@ <h2 id="toc_title" class="toc-title">Categories</h2>
...
+    </p>
+    <p>
+      In HTML5, several of the landmark roles are implicit via the native structural elements which are supported by modern <a href="http://www.html5accessibility.com/">desktop browsers</a>. With that said, if you need to support older browsers, you'll want to check their support for landmark roles. If you find support to be lacking, then use the landmark roles regardless of the validation warnings. For more information, read <a href="/posts/aria-landmark-roles/">Quick Tip: ARIA Landmark Roles and HTML5 Implicit Mapping</a>.
```
svinkle: > check their support for landmark roles

Is there a handy link we can share where folks could check this?

checklist.html
```diff
@@ -41,21 +48,23 @@ <h2 id="toc_title" class="toc-title">Categories</h2>
...
+		<p id="main-role-description" class="description">Wraps the focal content of document. Use only once.</p>
 
 		<!-- article -->
 		<label for="article-role" class="checkbox"><code>&lt;article role="article"&gt;</code>
 			<input name="aria-article-role" id="article-role" aria-describedby="article-role-description" type="checkbox">
 		</label>
 
-		<p id="article-role-description" class="description">Represents an independent item of content. Use only once on outermost element of this type.</p>
+		<p id="article-role-description" class="description">
+      Represents an independent item of content. There may be many &lt;article&gt;s in a single document. &lt;article&gt;s are not considered landmarks, but screen readers may still surface <code>article</code>s when navigating by regions or landmarks in a document.
```
svinkle: Let's be consistent when writing about code or elements of code. 

If we're discussing an HTML element or attribute name, wrap with `<code>` for formatting. I personally don't care for adding `&lt;` or `&gt;` as these are awkwardly spoken aloud by screen readers.

svinkle:
Thanks, @scottaohara. I left a few minor notes for you.

checklist.html
scottaohara: yup.  was thinking about that last night, but clearly got distracted / didn't change

checklist.html
scottaohara: Outside of the already linked HTML5accessibility.  I'm not aware of one.

checklist.html
scottaohara: agreed.  removed all instances of &lt; and &gt; (outside of once instance where there were multiple elements being referenced).  updating instances of elements not being wrapped in `code` elements to be.

Issue #501, opened 2017-07-25
Pattern: Accessible autocomplete


svinkle:
This autocomplete/combobox is pretty well done. https://github.com/alphagov/accessible-autocomplete (Examples: https://alphagov.github.io/accessible-autocomplete/examples/)

How do folks feel about linking to a GitHub repo instead of a CodePen? I know it's not ideal as there's more digging; perhaps this could be linked elsewhere on the site?

scottaohara:
my 2 cents, but I'd welcome patterns that linked to pages other than codepen, since it requires a pattern to be ported to codepen, which could then make it out of date with a github repo where it's actually maintained. 

Also testing pens, with screen readers, can be a bit annoying if one doesn't switch over to the full screen mode.

svinkle:
Right, I wouldn't expect someone create a new pen based on the repo, that'd be a maintenance nightmare. It's just that this method requires a bit more effort to find the code/demo in question, depending on which resource we link to.

I agree with the sentiment that it doesn't _have_ to be CodePen; any sort of code demo site would suffice.

svinkle:
Closing this as I'll probably make my own example to contribute, someday.

PR #583, opened 2018-02-01
add pattern: "dragon drop" accessible drag & drop

caseyg:
code by @schne324

demo: http://schne324.github.io/dragon-drop/demo/

via https://twitter.com/ted_drake/status/958509480991272960 (via @ablerism)

more info here: https://www.smashingmagazine.com/2018/01/dragon-drop-accessible-list-reordering/


svinkle:
We've had some reserve (#501) on linking directly to GitHub repos for patterns. I think this would be ok if it linked to the demo page, since there's a link to the source from there an a bit of code showing how to use the component.

@caseyg Could you update to link to the demo page instead.

PR #601, opened 2018-05-30
Hover and active states

ericwbailey:
This PR addresses https://github.com/a11yproject/a11yproject.com/issues/600. It adds hover and active states for links and buttons.

### DIFF
--- a/_sass/_colors.scss
+++ b/_sass/_colors.scss
@@ -1,6 +1,7 @@
 $grayLighter: #f0f0f0;
 $grayDarker: darken($grayLighter, 10%);
 $blue: #397c9c;
+$umber: #786017;
 
 $body_bg: #fdfdfd;
 $headings: #333;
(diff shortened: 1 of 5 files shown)


ericwbailey:
Here's a screencap of it in action:

Animation demonstrating the hover, active, and focused states for links. Hovered links remove the underline, active has a color change from dark green to umber and restore the underline, and focused states use the browser's native focus ring.

`focus` uses the browser focus ring for now, but I could whip up a background color shift or the like.

svinkle:
Looking good! Is that a custom `:focus` state? Or rather, which browser?

ericwbailey:
Ah, it's Chrome's default focus ring.

svinkle:
Ah I see what's happening; you've set your macOS appearance to "Graphite." 😄

LGTM 👍🚢

Issue #629, opened 2018-08-18
Add a style guide for content


missmatsuko:
I think it would be helpful to have a style guide for content contributions to keep content nice and consistent.

I noticed that color/colour is spelled different ways on the website, and even within the same article:
https://a11yproject.com/posts/understanding-colourblindness/

I also read this comment on a PR that mentions some conventions that I couldn't find formally recorded in a single file:
https://github.com/a11yproject/a11yproject.com/issues/516#issuecomment-324784426

ericwbailey:
I also think this is a good idea, thank you! If you get an initial PR ready, say a Markdown file in the repo's root, we can have a discussion about it there.

ericwbailey:
Yeah. I think it'd be better to point to CONTENT_STYLE_GUIDE.md from CONTRIBUTING.md. I'd prefer to segment it into "here's how to behave when you want to work with us," and "here's what to know when you write for us." It'd also make for a shorter contributing doc, which I think would help people actually reading it!

As for the guide, I think that's a great starting point! I'm also a big fan of [Shopify's](https://polaris.shopify.com/content/product-content#navigation).

ericwbailey:
Closed via https://github.com/a11yproject/a11yproject.com/pull/640

PR #632, opened 2018-08-18
Remove 'Estimated Time' and references to time

### LINKED ISSUE #630: Question: What does the 'Estimated Time' in Quick Test articles even mean?
I think it's supposed to be the time it takes to run the tests in the article, but wouldn't that be entirely dependant on the size of the site that's being tested? For example a short one-page landing page would take a lot less time to test than a 100+ page corporate website.

I also thought maybe it's the time it takes to read through the article and its references, but then why isn't that on all articles?

I would suggest removing the times, or clarifying what it means.

missmatsuko:
Closes #630

### DIFF
--- a/_posts/2013-01-11-navigate-using-just-your-keyboard.md
+++ b/_posts/2013-01-11-navigate-using-just-your-keyboard.md
@@ -2,13 +2,12 @@
 layout: post
 title: |
     Quick Test: Navigate using just your keyboard
-description: A quick 5 second accessibility test that you can run on your sites right now.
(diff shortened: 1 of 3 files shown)


scottaohara:
PR looks good and I think this makes sense.

If anyone else agrees, please merge.

ericwbailey:
Looks good to me as well, conversation here: https://github.com/a11yproject/a11yproject.com/issues/629

Merging this. Thank you for your contribution, @missmatsuko!

Issue #628, opened 2018-08-11
Question: Is the order of the categories intentional?


ptrin:
Hi, 

The order of categories in the navigation and on the homepage is:

**Categories**

- How-tos
- Myths
- Quick tests
- Quick tips
- Assistive technology
- Basics

Is there a reason for this order? Should we consider starting with Basics for people who are new to accessibility?

jeryj:
The original intention (right or wrong) of the homepage was to have a dev or designer be able to learn something they could put into practice within the first minute of being on the site. I think the Basics section would be a better fit for getting someone in for the long haul of learning web accessibility, and Quick Tips and How-Tos would be better for getting someone to implement _something_ and then learn more later.

Looking at the Basics section, it may serve from a rename. Maybe "Background"? I definitely agree that the order seems odd when looking at it in that category list :)

ericwbailey:
Ah, that makes sense! I don't think we'd be risking breaking too much if we renamed, considering the URLs for that section are `/posts/`.

svinkle:
Reading over the topics of what's under this section, I feel like these posts are more "advanced" or a "deep-dive" into a very specific topic. I'm on board with a rename as I wouldn't consider these "basic" concepts.

scottaohara:
Yeh, looking at each, i'm in the middle.  they're more intermediate topics, but I don't think i'd call them that?  Each is more than a quick tip, but none of them provide a full narrative on the topic.

ericwbailey:
Closed via #634

PR #634, opened 2018-08-20
Rename Basics category to Background

ericwbailey:
I'm still a little on the fence about "Background" as a term, I think it is better than the current category name, in that it addresses the issues outlined in https://github.com/a11yproject/a11yproject.com/issues/628


scottaohara:
I agree that we might come up with a better term for this, but for now I think this is good to merge while we noodle some more.

Issue #617, opened 2018-07-22
Add authors & updaters to all applicable posts


scottaohara:
There are a bunch of posts on the site that do not have authors, or attribution to those who have updated an article, who were not the original author.

We should review previous posts and 
1. Note the author (may require reviewing commit history)
2. Note the updater (if different from the original author)

Question:
Should we add a link to an author's name to have a link back to their websites / social media accounts?

ericwbailey:
I'm always a fan of attribution, so I'm a fan of this.

I know this balloons the effort involved, but we might want to reach out and confirm with the author/editor if they want that info added.

svinkle:
Wasn't this a recent PR?

scottaohara:
No, but I spoke to you about it in the context of another PR.

ericwbailey:
Asked for permission for some people whose contact info they made available. I figure I'll start adding it in if I get replies.

ericwbailey:
Here's the first pass: https://github.com/a11yproject/a11yproject.com/pull/638

PR #638, opened 2018-08-23
Attribution

ericwbailey:
This PR is a fist pass at adding a system for managing authors on the site. See https://github.com/a11yproject/a11yproject.com/issues/617

It will list the author's preferred name when displaying a post they have written, as well as a contact URL, if provided. 

I will be following this PR up with adding authors to posts without author information, after a little Git spelunking.

Preview: 

Published on January 14, 2013 by Erik Runyon - Updated: July 22, 2018 by Scott O'Hara. Scott's name is a hyperlink. Screenshot.

### DIFF
--- a/_data/authors.yml
+++ b/_data/authors.yml
@@ -13,6 +13,9 @@ erik_runyon:
 franklyn_roth:
   name: Franklyn Roth
   contact: https://www.linkedin.com/in/franklynroth
+gray_ghost:
+  name: Gray Ghost Visuals
+  contact: http://grayghostvisuals.com/
 greg_tarnoff:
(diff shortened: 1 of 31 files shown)


scottaohara:
initial review looks good, but haven't run the branch locally yet.

will get on that later.

ericwbailey:
Just pushed up attribution for the remaining published articles for authors that have gotten back to me.

ericwbailey:
I also standardized the way we treat font matter for things like post titles and descriptions, including how we case titles. I figure it'll be easier for a new contributor to get started if they don't have questions about how we format things.

Issue #603, opened 2018-06-13
Article: MYTHBUSTED: Accessibility is just turning off CSS


ValorNaram:
The post on https://github.com/a11yproject/a11yproject.com/issues/12 inspired me to write a short article for you to publish on your site.

Note: I am not good at writing articles and this is probably my first one, so I am open for better ones, corrections, additions etc.

You can check out my gist at https://gist.github.com/ValorNaram/161d5afdc5db1212ae81fb3531ab56d2

ValorNaram:
I updated my gist, please review the updated version and give feedback. In my article I also cover the problem we frontend web developers have with accessibility.

svinkle:
This is quite difficult to read. I would suggest running the content through https://hemingwayapp.com/ as a starting point.

PR #643, opened 2018-09-09
Article: Myth: Accessibility is just turning off CSS

ValorNaram:
Referring to issue [#603](https://github.com/a11yproject/a11yproject.com/issues/603). An article about `Myth: Accessibility is just turning off CSS`

### DIFF
--- a/_posts/myth-accessibility-is-just-turning-off-CSS.md
+++ b/_posts/myth-accessibility-is-just-turning-off-CSS.md
@@ -1,0 +1,32 @@
+---
+layout: post
+title: Myth: Accessibility is just turning off CSS
+description: Also users with disabilities need CSS
+author: [Valor Naram](https://github.com/valornaram)
+published: true
+categories:
+  - Myths
+---
+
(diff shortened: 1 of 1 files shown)


ValorNaram:
_By turning off CSS you forget that blind people are different. Some can distinct between light or dark, others can see structures and the other group of blinds can see a few metres. The structure seeing blinds may use the structure of a page to navigate even more fast. And imaging you turn off CSS for all these different blind people? The result: The structure seeing blinds may get stuck with this new situation._

I think this is a bit too complicated to understand. Any suggestions? I will think over it to find a way to explain it better, but I don't think that I will success

smythp:
Ho is this?

> By turning off CSS,  you are likely to be leaving out many individuals with visual impairments. Remember that blindness and visual impairment is a spectrum. Many, and in fact most, blind or visually impaired individuals have usable vision. By turning off CSS, you remove a tool that might be used by those with low vision or other visual impairments short of total blindness.

I'm low vision myself so this sentiment makes sense to me.

ValorNaram:
Sounds good. The way how I labeled the different groups in the spectrum was the problem.

ValorNaram:
Since I deleted my fork due to inactivity I need to close that PR and fork this repo again, doing the changes and then make a PR again.

PR #649, opened 2018-10-15
DL Style Changes

alexroseb:
When I was reviewing the site, the definition lists just turned into plaintext. This fix will make it more obvious to sighted users when there is a `<dl>`.

### DIFF
--- a/_sass/_reset.scss
+++ b/_sass/_reset.scss
@@ -65,6 +65,13 @@ a img {
     border: 0;
 }
 
+dd{
+    padding-left: 1em;
+}
+dt{
+    font-weight: bold; 
+}
+
 html {
     box-sizing: border-box;
 }


_sass/_reset.scss
```diff
@@ -65,6 +65,13 @@ a img {
     border: 0;
 }
 
+dd{
```
scottaohara: Might it be better to find where the `dl` & `dt` styles are being unset and remove those lines / update as needed?  

Additionally, I personally quite like this idea, but are there any pages on the site where this styling might muck something up?  (i would hope not, but we should be sure.)

Thanks!

alexroseb: They're unset a few lines above in this file, but I didn't want to change that as it also sets some other general properties for them. 

`dl`s are used in a couple articles where this will improve their look.
The only other place with a `dl` is [events.html](https://github.com/a11yproject/a11yproject.com/blob/gh-pages/events.html) to assist with screenreader compatibility, but it has class names on the `dt` and `dd`s, which should override this setting.

ericwbailey:
I'm happy with it. Merge it!

PR #655, opened 2018-10-26
clarified language

emilane:


ericwbailey:
Much better, thank you!

Issue #669, opened 2018-11-22
Review A11y Project's hosted content


scottaohara:
In an effort to ensure that The A11y Project's content is up-to-date, each hosted resource on the site should be reviewed to ensure code patterns are still relevant, links continue to work, and techniques are referencing the latest best practices.

If an article has been identified as needing an update, a new issue should be created and logged under the [Content Updates for 2019 Milestone](https://github.com/a11yproject/a11yproject.com/milestone/3) to track progress / request help from the community for updates.

If an article has recently been written, or received an update, then it can be checked off without further action.

If an article remains perfectly fine as-is, then it can be checked off without further action.

Assigning myself and @ericwbailey (sorry not sorry) as initial reviewers, but this issue is open to anyone who would like to participate.

The content to review is broken up by its parent topic grouping:

## How-tos
- [x] [Using skip navigation links](https://a11yproject.com/posts/skip-nav-links/)
- [x] [Create accessible data tables](https://a11yproject.com/posts/accessible-data-tables/)
- [x] [Using placeholder attributes](https://a11yproject.com/posts/placeholder-input-elements/)
- [x] [Hide content](https://a11yproject.com/posts/how-to-hide-content/)
- [x] [Using `title` attributes](https://a11yproject.com/posts/title-attributes/)
- [x] [Future proof your accessibility efforts.](https://a11yproject.com/posts/future-proof-your-accessibility/)
- [x] [Using caption services with HTML5 video](https://a11yproject.com/posts/using-caption-services-with-html5-video/)
- [x] [Using `role='application'`](https://a11yproject.com/posts/how-to-use-application-role/)

## Myths
- [x] [Screen readers don't use JavaScript](https://a11yproject.com/posts/myth-screen-readers-dont-use-javascript/)
- [x] [Accessibility is 'blind people'](https://a11yproject.com/posts/myth-accessibility-is-blind-people/)

## Quick tests
- [x] [Large touch targets](https://a11yproject.com/posts/large-touch-targets/)
- [x] [Check comprehension level](https://a11yproject.com/posts/check-comprehension-level/)
- [x] [Check contrast with your mobile device](https://a11yproject.com/posts/check-contrast-with-mobile-device/)
- [x] [Navigate using just your keyboard](https://a11yproject.com/posts/navigate-using-just-your-keyboard/)

## Quick tips
- [x] [Safari keyboard navigation](https://a11yproject.com/posts/safari-keyboard-navigation/)
[...]

PR #673, opened 2018-11-23
Myth: Accessibility is blind people article updates

ericwbailey:
This PR addresses the audit efforts in https://github.com/a11yproject/a11yproject.com/issues/669.

It converts the table into a series of subheadings. We currently do not support responsive tables, making the content difficult to read on at the mobile breakpoint. Additionally, the table may not be the best choice for this content, from both a semantic and SR-friendly perspective.

Additionally, it removes the last paragraph, as we haven't covered these categories further since the article's 2013 publish date.


ericwbailey:
Looks good to me, thanks!

PR #706, opened 2018-12-09
Updating articles to use further_reading front matter property

jacobpetersen:
Putting all "further reading" links from existing articles into the new further_reading front matter property


ericwbailey:
Go for it!

PR #758, opened 2019-03-30
Add contributing guidelines for promotions, partnership deals, and SEO scams 

ericwbailey:
We've had a rash of these recently, and I'd like to get some language added that speaks to it directly. The idea is while we can't prevent this from happening, we can have a concrete section we can link to when rejecting submissions.

If the language sounds harsh, it's intentional. I want to remove the potential for pushback for perceived ambiguity.


svinkle:
Agreed, good stuff. Thanks, Eric.

PR #762, opened 2019-04-10
When the class is used in an a, input or button element directly the element is not shown when receives focus.

luisortizramos:
When the class is used in an a, input or button element directly the element is not shown when receives focus. I have had the `.visually-hidden:focus` selector to the rule to fix this issue.


scottaohara:
Good catch @luisortizramos, unfortunately the issue goes deeper than that.   Giving this a quick test with even the updated selector chain, the [visually hidden content isn't revealed at all](https://codepen.io/scottohara/pen/gymXoe) per the CSS outlined in this article.

This advice should be changed to:
```css
.visually-hidden:not(:focus):not(:active) { 
  /* ... */
}
```

as that will actually work in ie9+ and negate the need to be so explicit with the selector chain.

Would you like to make another update to the post to correct this information? :)

Issue #736, opened 2019-01-31
Question: Comprehensiveness of Web Accessibility Checklist


nas5w:
Hi all, fantastic work on the a11yproject.com site and I look forward to the redesign! 

I was browsing the [Web Accessibility Checklist](https://a11yproject.com/checklist) and noted that, while everything on it seems really sound and in accordance with WCAG guidance, there are potentially missing items. For example, one item I would expect to see is in reference to ficker/flash/blink rate of any element. In the past, I have used checklists like the [webaim.org checklist](https://webaim.org/standards/wcag/checklist).

Of course, presenting a checklist with 100 items on it could easily scare off a newcomer. So my question is: how comprehensive is the [Web Accessibility Checklist](https://a11yproject.com/checklist) intended to be? Are you open to adding more items? Some of the more important items that I think might be missing involve:

- [Content linearization](https://webaim.org/techniques/screenreader/#linearization). You already have this in the forms section but really it applies to the entire application.
- [Instructions shouldn't depend on visual or audio](https://www.w3.org/TR/WCAG21/#sensory-characteristics). E.g., "click on the square to continue."
- [Color is not the only visual means of conveying information](https://www.w3.org/TR/WCAG21/#use-of-color). E.g., Red shouldn't mean "danger" without some accompanying text.
- [Pause Stop Hide](https://www.w3.org/TR/WCAG21/#pause-stop-hide) and [Three flashes or below](https://www.w3.org/TR/WCAG21/#three-flashes-or-below-threshold)
- [Images of text](https://www.w3.org/TR/WCAG21/#images-of-text). If text can convey the text in an image, consider just using the text.
- [No parsing errors](https://www.w3.org/TR/WCAG21/#parsing)

Thank you!

scottaohara:
Hey @nas5w, thanks for opening the issue!

The checklist has actually been a topic we've started talking about, and definitely will be getting some revising with the current website redesign project.  So this is definitely a good time to be raising ideas for improving it.

cc @svinkle @ericwbailey

nas5w:
Great, thank you! The WCAG 2.1 guideline is huge, so it's hard to envision a comprehensive checklist that isn't also huge. One ambitious idea I've had is to have a checklist generator of sorts: answer questions about your application and a checklist with only applicable items is generated. Not sure how that would work in practice!

sebacicltd:
I'm new to A11y, and I'm coming to it from a language perspective, but I'm
[...]

PR #779, opened 2019-04-28
Update checklist

ericwbailey:
This PR:

- Removes the form. Checking off checklist items does not create a situation where data will be submitted—if we want a persistent checked state, we should use a more modern solution such as `localStorage`.
    - Removes [garlic.js](https://garlicjs.org/) to reflect this.
- Adds some context about why the checklist exists, as well as links to resources to use to test. 
- Most importantly, there is a significant content update to the checklist items, reflecting @svinkle's efforts to better steer people towards AA compliance. Each checklist item now has a pertinent success criterion, description, and links where appropriate. The checklist sections have been expanded, as well.

This PR won't interfere with redesign efforts, as the checklist's place in the site's information architecture won't be moving. 

I know it's a good amount of content to review, but I'd personally love it if we could get this updated by [Global Accessibility Awareness Day](https://globalaccessibilityawarenessday.org/) (May 16th). I also know some people have been having difficulty building the site, so [here's a screenshot](https://user-images.githubusercontent.com/634191/56867560-7e916b80-69b4-11e9-82c0-17f831b82058.png).

Also tapping @nas5w and @sebacicltd, for concerns expressed in https://github.com/a11yproject/a11yproject.com/issues/736.

### DIFF
--- a/checklist.html
+++ b/checklist.html
@@ -540,13 +540,13 @@ permalink: checklist.html
         </p>
       </div>
     </details>
-    <details id="identify-links-that-open-in-a-new-window-or-links-that-navigate-to-external-sites">
+    <details id="identify-links-that-open-in-a-new-window">
       <summary>
-        Identify links that open in a new window, or links that navigate to external sites.
+        Identify links that open in a new window.
       </summary>
       <div class="checklist-details">
         <p class="checklist-criterion">
-          <a href="https://www.w3.org/TR/UNDERSTANDING-WCAG20/consistent-behavior-unpredictable-change.html">3.2.2 On Input</a>
+          <a href="https://www.w3.org/TR/WCAG20-TECHS/G201.html">G201: Giving users advanced warning when opening a new window</a>
         </p>
         <p class="checklist-description">
           Doing this will help people understand what will happen when activating the link. Ideally this information should be communicated in a way that is available to all users.
@@ -732,7 +732,7 @@ permalink: checklist.html
           <a href="https://www.w3.org/TR/UNDERSTANDING-WCAG20/keyboard-operation-keyboard-operable.html">2.1.1 Keyboard</a>
         </p>
         <p class="checklist-description">
-          Provide a global pause function on any media control. If the device has a keyboard, ensure that pressing the <kbd>Space</kbd> key can pause playback. Make sure you also don't interfere with the <kbd>Space</kbd> key's ability to scroll the page/view when not focusing on a form control.
+          Provide a global pause function on any media element. If the device has a keyboard, ensure that pressing the <kbd>Space</kbd> key can pause playback. Make sure you also don't interfere with the <kbd>Space</kbd> key's ability to scroll the page/view when not focusing on a form control.
         </p>
       </div>
     </details>
--- a/checklist.html
+++ b/checklist.html
@@ -28,32 +28,32 @@ permalink: checklist.html
 <section class="resources-section checklist-introduction" id="section-introduction" aria-labelledby="label-introduction">
   <h2 id="label-introduction">Introduction</h2>
   <p>
-    This checklist uses the <a href="https://www.w3.org/WAI/standards-guidelines/wcag/">The Web Content Accessibility Guidelines (<abbr>WCAG</abbr>)</a>, a shared standard for web content accessibility for individuals, organizations, and governments. There are three levels of compliance which reflect the priority of accessibility support:
+    This checklist uses the <a href="https://www.w3.org/WAI/standards-guidelines/wcag/">The Web Content Accessibility Guidelines (<abbr>WCAG</abbr>)</a> as a reference point. The WCAG is a shared standard for web content accessibility for individuals, organizations, and governments.
+  </p>
+  <p>
+    There are three levels of accessibility compliance in the WCAG, which reflect the priority of support:
   </p>
   <ol>
     <li><strong>A</strong>: Essential. If this isn't met, assistive technology cannot read, understand, or operate the technology.</li>
     <li><strong>AA</strong>: Ideal support. Now required for EU government and public body websites.</li>
-    <li><strong>AAA</strong>: High level support. This is typically reserved for sites that serve a specialized audience.</li>
+    <li><strong>AAA</strong>: High-level support. This is typically reserved for sites that serve a specialized audience.</li>
   </ol>
   <p>
-    Note that the different levels of WCAG support do not necessarily indicate an increased level of difficulty to implement. This checklist is built using WCAG as a reference point. It targets many, but not all level AA issues.
+    This checklist targets many, but not all level AA concerns. Note that the different levels of WCAG support do not necessarily indicate an increased level of difficulty to implement.
   </p>
   <h3 id="success-criteria">Success criteria</h3>
   <p>
-    Each item on this checklist corresponds to a "success criterion," which is a specific rule in the WCAG. Each success criterion has a number and a short title. For example, the rule about text resizing is called <a href="https://www.w3.org/TR/UNDERSTANDING-WCAG20/visual-audio-contrast-scale.html">1.4.4 Resize text</a>.
+    Each item on this checklist has a corresponding WCAG "success criterion." Success criterion are the specific, testable rules that power the WCAG, described by a reference number and short title. For example, the rule about text resizing is called <a href="https://www.w3.org/TR/UNDERSTANDING-WCAG20/visual-audio-contrast-scale.html">1.4.4 Resize text</a>.
   </p>
   <p>
     Some accessibility issues may have multiple success criterion apply to them. We have identified the most relevant for each checklist item.
   </p>
   <h3 id="does-this-checklist-guarantee-my-site-is-accessible">Does this checklist guarantee my site is accessible?</h3>
   <p>
-    No. However, using this checklist will help improve the experience for everyone who uses your site. The issues it prompts you to check for covers a wide range of disability conditions.
+    No. However, going through and fixing issues for every item this checklist will help greatly improve the experience for everyone who uses your site.
   </p>
   <p>
-    Using this checklist will help make sure your site stays flexible and fault-tolerant. There is no such thing as "perfect accessibility" or a site being "100% accessible". You should be wary of companies and services that promise this.
-  </p>
-  <p>
-    Need professional accessibility help? Use <a href="resources#professional-help">professional accessibility services</a>.
+    The issues this checklist prompts you to check for covers a wide range of disability conditions. There is no such thing as "perfect accessibility" or a site being "100% accessible". You should be wary of companies and services that promise this. If you need professional accessibility help, use <a href="resources#professional-help">professional accessibility services</a>.
   </p>
 </section>
 
--- a/checklist.html
+++ b/checklist.html
@@ -34,12 +34,12 @@ permalink: checklist.html
     There are three levels of accessibility compliance in the WCAG, which reflect the priority of support:
   </p>
   <ol>
-    <li><strong>A</strong>: Essential. If this isn't met, assistive technology cannot read, understand, or operate the technology.</li>
+    <li><strong>A</strong>: Essential. If this isn't met, assistive technology may not be able to read, understand, or fully operate the page or view.</li>
     <li><strong>AA</strong>: Ideal support. Now required for EU government and public body websites.</li>
     <li><strong>AAA</strong>: High-level support. This is typically reserved for sites that serve a specialized audience.</li>
   </ol>
   <p>
-    This checklist targets many, but not all level AA concerns. Note that the different levels of WCAG support do not necessarily indicate an increased level of difficulty to implement.
+    This checklist targets many, but not all level A and AA concerns. Note that the different levels of WCAG support do not necessarily indicate an increased level of difficulty to implement.
   </p>
   <h3 id="success-criteria">Success criteria</h3>
   <p>
@@ -50,10 +50,10 @@ permalink: checklist.html
   </p>
   <h3 id="does-this-checklist-guarantee-my-site-is-accessible">Does this checklist guarantee my site is accessible?</h3>
   <p>
-    No. However, going through and fixing issues for every item this checklist will help greatly improve the experience for everyone who uses your site.
+    No. However, addressing the issues called out in this checklist will help improve the experience for everyone who uses your site.
   </p>
   <p>
-    The issues this checklist prompts you to check for covers a wide range of disability conditions. There is no such thing as "perfect accessibility" or a site being "100% accessible". You should be wary of companies and services that promise this. If you need professional accessibility help, use <a href="resources#professional-help">professional accessibility services</a>.
+    The issues this checklist prompts you to check for covers a wide range of disability conditions. There is no such thing as "perfect accessibility" or a site being "100% accessible". You should be wary of companies and services that make such promises. If you need professional accessibility help, use <a href="resources#professional-help">professional accessibility services</a>.
   </p>
 </section>
 
@@ -63,7 +63,7 @@ permalink: checklist.html
     Content is the most important part of your site.
   </p>
   <div class="checklist">
-    <details id="section-content-01">
+    <details id="use-plain-language-and-avoid-figures-idioms-and-complicated-metaphors">
       <summary>
         Use plain language and avoid figures of speech, idioms, and complicated metaphors.
       </summary>
@@ -76,7 +76,7 @@ permalink: checklist.html
         </p>
       </div>
     </details>
-    <details id="section-content-02">
+    <details id="make-sure-that-button-a-and-label-element-content-is-unique-and-descriptive">
       <summary>
         Make sure that <code>button</code>, <code>a</code>, and <code>label</code> element content is unique and descriptive.
       </summary>
@@ -89,7 +89,7 @@ permalink: checklist.html
         </p>
       </div>
     </details>
-    <details id="section-content-03">
+    <details id="use-left-aligned-text-for-left-to-right-languages-and-right-aligned-text-for-right-to-left-languages">
       <summary>
         Use left-aligned text for left-to-right languages, and right-aligned text for right-to-left languages.
       </summary>
@@ -111,20 +111,20 @@ permalink: checklist.html
     Global code is code that affects your entire website.
   </p>
   <div class="checklist">
-    <details id="section-global-code-01">
+    <details id="validate-your-html">
       <summary>
-        Validate the HTML you generate.
+        Validate your HTML.
       </summary>
       <div class="checklist-details">
         <p class="checklist-criterion">
           <a href="https://www.w3.org/TR/UNDERSTANDING-WCAG20/ensure-compat-parses.html">4.1.1 Parsing</a>
         </p>
         <p class="checklist-description">
-          <a href="https://validator.w3.org/nu/">Valid HTML</a> helps give a consistent experience across all browsers and assistive technology.
+          <a href="https://validator.w3.org/nu/">Valid HTML</a> helps to provide an consistent, expected experience across all browsers and assistive technology.
         </p>
       </div>
     </details>
-    <details id="section-global-code-02">
+    <details id="use-a-lang-attribute-on-the-html-element">
       <summary>
         Use a <code>lang</code> attribute on the <code>html</code> element.
       </summary>
@@ -137,7 +137,7 @@ permalink: checklist.html
         </p>
       </div>
     </details>
-    <details id="section-global-code-03">
+    <details id="provide-a-unique-title-for-each-page-or-view">
       <summary>
         Provide a unique <code>title</code> for each page or view.
       </summary>
@@ -146,11 +146,11 @@ permalink: checklist.html
           <a href="https://www.w3.org/TR/UNDERSTANDING-WCAG20/navigation-mechanisms-title.html">2.4.2 Page Titled</a>
         </p>
         <p class="checklist-description">
-          The <code>title</code> element, contained in the document's <code>head</code> element, is often the first piece of information announced by assistive technology. This helps tell people what screen or view they are going to start navigating.
+          The <code>title</code> element, contained in the document's <code>head</code> element, is often the first piece of information announced by assistive technology. This helps tell people what page or view they are going to start navigating.
         </p>
       </div>
     </details>
-    <details id="section-global-code-04">
+    <details id="ensure-that-viewport-zoom-is-not-disabled">
       <summary>
         Ensure that viewport zoom is not disabled.
       </summary>
@@ -159,24 +159,24 @@ permalink: checklist.html
           <a href="https://www.w3.org/TR/UNDERSTANDING-WCAG20/visual-audio-contrast-scale.html">1.4.4 Resize text</a>
         </p>
         <p class="checklist-description">
-          Some people need to increase the size of text to a point where they can read it. Do not stop them from doing this, even for sites with a native app-like experience.
+          Some people need to increase the size of text to a point where they can read it. Do not stop them from doing this, even for sites with a native app-like experience (even native apps should respect Operating System settings for resizing text).
         </p>
       </div>
     </details>
-    <details id="section-global-code-05">
+    <details id="use-landmark-elements-to-indicate-important-content-regions">
       <summary>
-        Use landmark elements to describe the large regions of content on a page.
+        Use landmark elements to indicate important content regions.
       </summary>
       <div class="checklist-details">
         <p class="checklist-criterion">
           <a href="https://www.w3.org/TR/UNDERSTANDING-WCAG20/ensure-compat-rsv.html">4.1.2 Name, Role, Value</a>
         </p>
         <p class="checklist-description">
-          Landmark regions help communicate the layout and important areas of a page, and can allow quick access to these regions. For example, use the <code>nav</code> element to wrap a site's navigation, or the <code>main</code> element to contain the primary content of a page.
+          Landmark regions help communicate the layout and important areas of a page or view, and can allow quick access to these regions. For example, use the <code>nav</code> element to wrap a site's navigation, or the <code>main</code> element to contain the primary content of a page.
         </p>
       </div>
     </details>
-    <details id="section-global-code-06">
+    <details id="ensure-a-linear-content-flow">
       <summary>
         Ensure a linear content flow.
       </summary>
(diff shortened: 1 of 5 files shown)


checklist.html
```diff
@@ -1,242 +1,428 @@
...
+  <h2 id="label-required-software">Required software</h2>
+  <p>
+    These programs can help you uncover accessibility issues in your code, or provide the functionality to determine if assistive technology can actually work with your website.
+  </p>
+  <h3 id="automated-testing-tools">Automated testing</h3>
+  <p>
+    Automated tests can check your code for common accessibility issues. Be sure to run <a href="https://www.smashingmagazine.com/2018/09/importance-manual-accessibility-testing/">two or more automated testing tools</a> to ensure comprehensive results, as one automated testing tool may not know to check for some kinds of code that another one does.
+  </p>
+  <ul>
+    <li><a href="https://www.deque.com/axe/">aXe</a>, by Deque</li>
```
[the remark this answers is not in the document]
svinkle: We could leave this section out, or link to the [Online Tools section](https://a11yproject.com/resources#online-tools).

svinkle:
Looks great, @ericwbailey! The extra content pieces at the top are a fantastic addition.

Curious if we should add some checkbox elements, to add that "checklist" feel?

checklist.html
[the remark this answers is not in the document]
scottaohara: Agreed @svinkle 

Eric and I spoke about this and the screen reader section as well.  good content, but not necessary for the checklist itself.

scottaohara:
thoughts on marking these up as a standard lists, but using details / summary (or ARIA disclosures) to show/hide the content?

I ask for a couple reasons:
* there’s a lot of additional content here now, which is good, but it doesn’t make for quick lists to review. 
* screen readers / browsers are inconsistent with the way they expose and let people navigate `dl`s and such, that while semantically i’m 100% on board with using these elements, practically i’ve yet to receive positive feedback from people who use screen readers, when implementing these elements.

re: the above two points, if checkboxes are used somehow, i'd expect they be in a visual manner.  It would be odd to have checkboxes and controls to show/hide.  

If we don't go with the disclosure widget route and do re-introduce checkboxes, let's just keep it simple.  There's no reason to reintroduce the `fieldset`s and the wrapping `form` element from the previous version.  This is an informative page, not actionable after all.

svinkle:
+1 native disclosure. Would save a ton of space and would help the content be less overwhelming which is sort of the point.

checklist.html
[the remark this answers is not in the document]
ericwbailey: Yup. I'll be gone with the next pass.

ericwbailey:
No problems here going with a native disclosure. I'll get on that soon.

ericwbailey:
Pushed up a new version with copy edits and a treatment using `details`. Here's a screencap:

Screenshot of the keyboard section, showing the collapsed and expanded states for each item.

sebacicltd:
I'm one of those people who can't build the site... so thanks for the
screengrab.

I've got a few comments I'd like to make about readability and
understandability - could someone cut and paste the whole page into a
google doc for these types of comments??


regards

Jonathan Holden

sebacicltd:
I'm also interested in WHO has responsibility for each of these "checkbox"
items during web design and build.

On a one person team, that's easy enough - but of course we need to "teach"
the client how to write good content. And we need to explain why we can't
use their logo colour for buttons (for example, contrast).

On a larger team, responsibility may fall to different people, and being
able to divide up the checklist according to role would be really helpful.

Clearly defining responsibility at the start of the project would help
execution a lot, I think.



regards

Jonathan Holden

scottaohara:
I like this idea, though I wonder how might be the best way to surface this, as many checklist items *should* be everyone's responsibility, to some degree.

sebacicltd:
Scott

I'm not so sure. The content writers have no responsibility for the skip
link. Then colour contrast of buttons and links is the responsibility of
the designer. Image alt tags? The system has to be set up to accept them
(ok, that's built in to WordPress for example) , and then it's down to the
person who uploads the images to tag them appropriately.

I feel that allowing the project manager to delegate specific tasks may
raise overall compliance as there will be less opportunity for people to
assume that "someone else will do that".

scottaohara:
@sebacicltd as stated "to some degree".  Per even your examples:

skip link requires a design and expected behavior (does it remain positioned, or does it become static?  how does this surface on small screens where iOS doesn't reveal visually hidden skip links on focus. developers then need to implement appropriately). 

colo(u)r contrast is initially a designers responsibility to provide, but dev and or QA *should* be empowered to verify implementing. Especially if using a component library where individual elements may be fine in isolation, but issues may become apparent in implementation.

Images being uploaded to a CMS may not be by the person who is in charge of how they'd be used.  e.g. interns doing the grunt work of optimizing for web and getting content uploaded may have no background in providing appropriate alt text.  Is the image used in one instance as decorative and another as informative?  Who makes that call and who can be on call to check so as to not have a single point of failure.

ericwbailey:
I'll get a Google doc with the content soon, @sebacicltd. Appreciate you taking a look.

Point of order: while we're discussing the responsibility topic, do we agree that it's not a blocker for merging the updated checklist content? I think we can start a new issue if and when this is merged, then potentially add in content from there.

sebacicltd:
Responsibility..... Absolutely not blocking anything.... Every project is
different, so it will be up to the project manager to decide in each case.

(it's just very pertinent to me personally at the moment)

ericwbailey:
Great. [Here's that content doc I promised](https://docs.google.com/document/d/1OJlANEkLp6-o6FcaE8zU8Zdzn1flzyvRfjpHr1zzxCU/edit?usp=sharing), it's currently set to read and suggest mode.

ericwbailey:
I'm still working through the doc, but these are some great edits so far. Thank you for doing them, @sebacicltd.

sebacicltd:
I'll take a look in the morning to see if you have any questions, but I'm
sure you know more about this than I do!

ericwbailey:
Just pushed up changes reflecting @sebacicltd's work.

checklist.html
```diff
@@ -288,42 +318,56 @@ <h3 id="svg-images">SVG Images</h3>
...
+        Ensure that <code>img</code> elements with an <code>svg</code> source includes the <code>role="img"</code> attribute.
       </summary>
       <div class="checklist-details">
-        <p class="checklist-criterion"><a href="https://www.w3.org/TR/UNDERSTANDING-WCAG20/ensure-compat-rsv.html">4.1.2 Name, Role, Value</a></p>
-        <p class="checklist-description">This code is used to convey image semantics correctly for VoiceOver on iOS.</p>
+        <p class="checklist-criterion">
+          <a href="https://www.w3.org/TR/UNDERSTANDING-WCAG20/ensure-compat-rsv.html">4.1.2 Name, Role, Value</a>
+        </p>
+        <p class="checklist-description">
+          This code is used to convey image semantics correctly for VoiceOver on iOS.
```
scottaohara: and macOS

checklist.html
```diff
@@ -332,44 +376,59 @@ <h3 id="svg-images">SVG Images</h3>
 <section class="article-section resources-section checklist-section" id="section-headings" aria-labelledby="label-headings">
   <h2 id="label-headings">Headings</h2>
   <p>
-    Headings elements (<code>h1</code>, <code>h2</code>, <code>h3</code>, etc.) help break up the content of the page into related "chunks" of information.
+    Heading elements (<code>h1</code>, <code>h2</code>, <code>h3</code>, etc.) help break up the content of the page into related "chunks" of information.
```
scottaohara: implicit sections of information

checklist.html
```diff
@@ -332,44 +376,59 @@ <h3 id="svg-images">SVG Images</h3>
...
   </p>
   <div class="checklist">
-
     <details id="section-headings-01">
       <summary>
-        Use headings elements to describe headings.
+        Use heading elements to describe headings.
```
scottaohara: Use heading elements to introduce content

checklist.html
```diff
@@ -396,44 +459,59 @@ <h2 id="label-lists">Lists</h2>
...
+        Provide a skip link and make sure that it is visible when focused.
       </summary>
       <div class="checklist-details">
-        <p class="checklist-criterion"><a href="https://www.w3.org/TR/UNDERSTANDING-WCAG20/navigation-mechanisms-skip.html">2.4.1 Bypass Blocks</a></p>
-        <p class="checklist-description"><a href="/posts/skip-nav-links/">Skip links</a> provide quick access the main page content, allowing a person to quickly and easily bypass repeated content such as a site's main navigation.</p>
+        <p class="checklist-criterion">
+          <a href="https://www.w3.org/TR/UNDERSTANDING-WCAG20/navigation-mechanisms-skip.html">2.4.1 Bypass Blocks</a>
+        </p>
+        <p class="checklist-description">
+          <a href="/posts/skip-nav-links/">Skip links</a> provide quick access to the main page content. This allows a person to quickly and easily bypass repeated content such as a site's main navigation.
```
scottaohara: A skip link can be used to provide quick access to the main content of a page or view. This will allow a person to easily bypass globally repeated content like a website's primary navigation, or persistent search widget.

checklist.html
```diff
@@ -396,44 +459,59 @@ <h2 id="label-lists">Lists</h2>
...
         Identify links that open in a new window, or links that navigate to external sites.
       </summary>
       <div class="checklist-details">
-        <p class="checklist-criterion"><a href="https://www.w3.org/TR/UNDERSTANDING-WCAG20/consistent-behavior-unpredictable-change.html">3.2.2 On Input</a></p>
-        <p class="checklist-description">Doing so will help people not be surprised by the result of activating the link.</p>
+        <p class="checklist-criterion">
+          <a href="https://www.w3.org/TR/UNDERSTANDING-WCAG20/consistent-behavior-unpredictable-change.html">3.2.2 On Input</a>
+        </p>
+        <p class="checklist-description">
+          Doing this will help people understand what will happen when activating the link.
```
scottaohara: Ideally this information should be communicated in a way that is available to all users.

checklist.html
```diff
@@ -482,68 +571,95 @@ <h2 id="label-forms">Forms</h2>
...
         Update the page or view's <code>title</code> to include high-level form feedback.
       </summary>
       <div class="checklist-details">
-        <p class="checklist-criterion"><a href="https://www.w3.org/TR/UNDERSTANDING-WCAG20/content-structure-separation-programmatic.html">1.3.1 Info and Relationships</a></p>
-        <p class="checklist-description">The <code>title</code> element's content is often the first content announced. Adding feedback about the presence of things like errors or warnings allows people to know there are items on the page or view that need their attention.</p>
+        <p class="checklist-criterion">
+          <a href="https://www.w3.org/TR/UNDERSTANDING-WCAG20/content-structure-separation-programmatic.html">1.3.1 Info and Relationships</a>
+        </p>
+        <p class="checklist-description">
+          The <code>title</code> element's content is often the first content announced by a screen reader. Adding feedback about the presence of things like errors or warnings allows people to know there are items on the page or view that need their attention.
```
scottaohara: screen readers won't announce a change to the content of a `title` element.  So this would be more helpful for a sighted user, or if a full page reload/refresh were to occur, which would then cause the screen reader to announce the title again.

checklist.html
```diff
@@ -28,70 +28,35 @@ <h2 id="toc_title" class="toc-title">Categories</h2>
...
+    <li><strong>A</strong>: Essential. If this isn't met, assistive technology cannot read, understand, or operate the technology.</li>
+    <li><strong>AA</strong>: Good support. Now required for EU government and public body websites.</li>
+    <li><strong>AAA</strong>: High level support. This is typically reserved for sites that serve a specialized audience.</li>
   </ol>
   <p>
-    While <strong>A</strong> support is mandatory, most websites should have a level of <strong>AA</strong> support to best guarantee support. <strong>AAA</strong> level compliance is typically reserved for sites that serve a specialized audience. It should also be noted that the different levels of WCAG support do not necessarily indicate an increased level of difficulty to implement.
-  </p>
-  <p>
-    This checklist is built using WCAG as a reference point. It targets many, but not all level <strong>AA</strong> issues.
+    Note that the different levels of WCAG support do not necessarily indicate an increased level of difficulty to implement. This checklist is built using WCAG as a reference point. It targets many, but not all level AA issues.
```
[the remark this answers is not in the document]
ericwbailey: So, "success criterion" is what the WCAG uses, but it's a jargon term that's kind of difficult to grok at first pass. Should we be consistent with honoring the WCAG terminology, or explain what it means and use our own terminology?

checklist.html
```diff
@@ -139,62 +116,90 @@ <h2 id="label-global-code">Global code</h2>
...
         Ensure a linear content flow.
       </summary>
       <div class="checklist-details">
-        <p class="checklist-criterion"><a href="https://www.w3.org/TR/UNDERSTANDING-WCAG20/navigation-mechanisms-focus-order.html">2.4.3 Focus Order</a></p>
-        <p class="checklist-description">Remove <code>tabindex</code> attribute values that aren't either <code>0</code> or <code>-1</code>. Remove the use of the <code>autofocus</code> attribute.</p>
+        <p class="checklist-criterion">
+          <a href="https://www.w3.org/TR/UNDERSTANDING-WCAG20/navigation-mechanisms-focus-order.html">2.4.3 Focus Order</a>
+        </p>
+        <p class="checklist-description">
+          Remove <code>tabindex</code> attribute values that aren't either <code>0</code> or <code>-1</code>. Remove the use of the <code>autofocus</code> attribute.
```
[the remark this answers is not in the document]
ericwbailey: How could we rephrase this? Developers slapping in `tabindex` everywhere, with a manually curated order and/or applying it to everything to make it "extra accessible" are both common issues, and I want to make sure we have something that addresses it.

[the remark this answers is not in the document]
ericwbailey: One thought: make `autofocus` its own checklist item to start.

checklist.html
ericwbailey: A lotta ten cent words in that phrase.

scottaohara: touche

checklist.html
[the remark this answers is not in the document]
scottaohara: Understood, but my comment was about the fact that right now "success criterion" is used in the prose as well as "issues" so we should be consistent.

Personally though, per your response, I'm more in favor of educating people to what the term means and then using it, rather than flat out shielding people from having to learn something.

checklist.html
ericwbailey: As an aside, we need to do this for our site.

checklist.html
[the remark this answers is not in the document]
ericwbailey: Pushed up some copy tweaks for the intro section. How does it read to you now?

ericwbailey:
Unless there's any more feedback, I will merge this on Saturday to provide some time for people to review it on the live site before we announce it on Wednesday.

checklist.html
```diff
@@ -1,242 +1,1029 @@
...
+    <details id="section-images-05">
+      <summary>
+        Make sure that <code>svg</code> elements include the code <code>focusable="false"</code> when they are the child element of a focusable element.
+      </summary>
+      <div class="checklist-details">
+        <p class="checklist-criterion">
+          <a href="https://www.w3.org/TR/UNDERSTANDING-WCAG20/content-structure-separation-programmatic.html">1.3.1 Info and Relationships</a>
+        </p>
+        <p class="checklist-description">
+          This prevents a Internet Explorer from allowing focus to navigate through the child elements of an SVG that is meant to be decorative.
```
svinkle: > This prevents <del>a</del> Internet Explorer

checklist.html
```diff
@@ -1,242 +1,1029 @@
...
+    <details id="section-images-07">
+      <summary>
+        Make sure that <code>svg</code> images which utilize the <code>use</code> element have whitespace between the <code>svg</code> and <code>use</code> elements.
+      </summary>
+      <div class="checklist-details">
+        <p class="checklist-criterion">
+          <a href="https://www.w3.org/TR/UNDERSTANDING-WCAG20/keyboard-operation-trapping.html">2.1.2 No Keyboard Trap</a>
+        </p>
+        <p class="checklist-description">
+          This solves a bug in Safari which interferes with VoiceOver describing <code>svg</code> images that contain the <code>use</code> element.
```
svinkle: > <del>interferes with VoiceOver describing <code>svg</code></del> can sometimes create hidden tab-stops when navigating

checklist.html
```diff
@@ -1,242 +1,1029 @@
...
+    <details id="section-forms-03">
+      <summary>
+        Inputs use <code>autocomplete</code> where appropriate.
+      </summary>
+      <div class="checklist-details">
+        <p class="checklist-criterion">
+          <a href="https://www.w3.org/WAI/WCAG21/Understanding/identify-input-purpose.html">1.3.5 Identify Input Purpose</a>
+        </p>
+        <p class="checklist-description">
+          <a href="https://www.w3.org/TR/html52/sec-forms.html#sec-autofill">Providing a mechanism</a> to help people more quickly, easily, and accurately fill in form fields that ask for common information (for example, name, address, phone number). Ensure that form input errors are displayed in a list above the form after the form is submit
```
svinkle: > <del>Ensure that form input errors are displayed in a list above the form after the form is submit ted.</del>

svinkle:
Two points I feel like are missing from the old list:

- Ensure links have a `:focus` state. 
- Ensure links are recognizable (underlined).

We could expand these to include `:active` state for visual confirmation (2.1.1 Keyboard), and to mention, at the very least, include underlines for link within body copy (1.4.1 Use of Color).

checklist.html
```diff
@@ -1,242 +1,1029 @@
...
+        </p>
+      </div>
+    </details>
+    <details id="section-controls-04">
+      <summary>
+        Identify links that open in a new window, or links that navigate to external sites.
+      </summary>
+      <div class="checklist-details">
+        <p class="checklist-criterion">
+          <a href="https://www.w3.org/TR/UNDERSTANDING-WCAG20/consistent-behavior-unpredictable-change.html">3.2.2 On Input</a>
```
scottaohara: Not doing this would not actually fail this SC.  more of an advisory technique / best practice, which is actually indicated in the understanding link, which then links to: https://www.w3.org/TR/2016/NOTE-WCAG20-TECHS-20161007/G201

checklist.html
```diff
@@ -1,242 +1,1029 @@
...
+
+<section class="article-section resources-section checklist-section" id="section-tables" aria-labelledby="label-tables">
+  <h2 id="label-tables">Tables</h2>
+  <p>
+    Tables are a structured set of data that help people understand the relationships between different types of information.
+  </p>
+  <div class="checklist">
+    <details id="section-tables-01">
+      <summary>
+        Use the <code>table</code> element to describe tabular data.
```
scottaohara: Use the <code>table</code> element to markup tabular data.

ericwbailey:
Thanks @svinkle. I've added them to the Controls section, starting here: https://github.com/a11yproject/a11yproject.com/pull/779/files#diff-f716bc61caf8adb552960a60746bd411R478

checklist.html
```diff
@@ -1,242 +1,1029 @@
...
+    <details id="section-forms-07">
+      <summary>
+        Update the page or view's <code>title</code> to include high-level form feedback.
+      </summary>
+      <div class="checklist-details">
+        <p class="checklist-criterion">
+          <a href="https://www.w3.org/TR/UNDERSTANDING-WCAG20/content-structure-separation-programmatic.html">1.3.1 Info and Relationships</a>
+        </p>
+        <p class="checklist-description">
+          The <code>title</code> element's content is often the first content announced by a screen reader. Adding feedback about the presence of things like errors or warnings after form submission allows people to know there are items on the page or view that need their attention.
```
scottaohara: Thought I had mentioned something about this before, but updating the `title` element will not be announced to screen readers.  this sort of advice is only applicable if a form submission completely refreshes the page (non-spa) and the user has to renavigate the document as if they've gone to a new page.

checklist.html
```diff
@@ -1,242 +1,1029 @@
...
+    <details id="section-media-02">
+      <summary>
+        Ensure that media controls use appropriate markup.
+      </summary>
+      <div class="checklist-details">
+        <p class="checklist-criterion">
+          <a href="https://www.w3.org/TR/UNDERSTANDING-WCAG20/content-structure-separation-programmatic.html">1.3.1 Info and Relationships</a>
+        </p>
+        <p class="checklist-description">
+          Examples include making sure a video play button has a pressed toggle state when active, or that a volume slider uses <code>&#x3C;input type=&#x22;range&#x22;&#x3E;</code>.
```
scottaohara: i'd swap out "play" for "mute" here.

play buttons more typically toggle their accessible name with play/pause, so a "toggle button" e.g. `aria-pressed=t/f` wouldn't be applicable here.

checklist.html
```diff
@@ -1,242 +1,1029 @@
...
+    <details id="section-media-03">
+      <summary>
+        Check to see that all media can be paused.
+      </summary>
+      <div class="checklist-details">
+        <p class="checklist-criterion">
+          <a href="https://www.w3.org/TR/UNDERSTANDING-WCAG20/keyboard-operation-keyboard-operable.html">2.1.1 Keyboard</a>
+        </p>
+        <p class="checklist-description">
+          Provide a global pause function on any media control. If the device has a keyboard, ensure that pressing the <kbd>Space</kbd> key can pause playback.
```
scottaohara: >Provide a global pause function on any media element that can be played.

we need to be careful with:
>ensure that pressing the <kbd>Space</kbd> key can pause playback.

Since space key should allow for the viewport to scroll, when not focusing a form control of some type, we don't want to make people think it's always ok to overwrite this functionality and disallow its inherent behavior.

typically the `space` key should allow for a "global pause" only when keyboard focus is within the context of a larger media grouping, and where space shouldn't also be expected to activate other interactive elements which are focused.

checklist.html
```diff
@@ -1,242 +1,1029 @@
...
+    <details id="section-content-02">
+      <summary>
+        Make sure that <code>button</code>, <code>a</code>, and <code>label</code> element content is unique and descriptive.
+      </summary>
+      <div class="checklist-details">
+        <p class="checklist-criterion">
+          <a href="https://www.w3.org/TR/UNDERSTANDING-WCAG20/content-structure-separation-programmatic.html">1.3.1 Info and Relationships</a>
+        </p>
+        <p class="checklist-description">
+          Terms like "click here" and "read more" do not provide any context. Some people navigate using a list of all buttons or links on a page or view. When using this mode, the terms indicate what will happen if navigated to or activated.
```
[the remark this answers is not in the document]
ericwbailey: I like this, but I'd also like to update it to include the cognitive angle, in that it's less ambiguous to all people, regardless of browsing mode.

checklist.html
```diff
@@ -1,242 +1,1029 @@
...
+    <details id="section-global-code-03">
+      <summary>
+        Provide a unique <code>title</code> for each page or view.
+      </summary>
+      <div class="checklist-details">
+        <p class="checklist-criterion">
+          <a href="https://www.w3.org/TR/UNDERSTANDING-WCAG20/navigation-mechanisms-title.html">2.4.2 Page Titled</a>
+        </p>
+        <p class="checklist-description">
+          The <code>title</code> element, contained in the document's <code>head</code> element, is often the first piece of information announced by assistive technology. This helps tell people what screen or view they are going to start navigating.
```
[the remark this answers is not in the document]
ericwbailey: We should settle on a term and add it to the writing styleguide.

I've been using "page/view" to specifically name both SPA and non-SPA use cases, as I think it'll preempt potential arguments about the whole "SPAS aren't documents!" crowd.

checklist.html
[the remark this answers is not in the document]
scottaohara: do it to it

checklist.html
svinkle: Agreed, though Léonie recently described her frustration with Play controls and the accessible name changing; she'd rather have the `aria-pressed` applied. 🤷‍♂️

https://youtu.be/OUDV1gqs9GA?t=3220

checklist.html
svinkle: This assumed full page refresh. It's still good practice, but folks will need to take other action for SPAs to convey the same thing.

checklist.html
scottaohara: Right, and from a completely visually impaired perspective that makes sense.  but as the visual icon of a play/pause often changes as well, it'd be just as confusing for other AT users if the accessible name remained play, but the button displayed a pause icon.

hence change the example to mute, since that's no need to change that label and it fits the pressed / not pressed example better.

checklist.html
scottaohara: We should explicitly call that out then since so many forms even outside of full SPAs could use JS to just update the current page, where this advice still wouldn't hold true without that clarity.

checklist.html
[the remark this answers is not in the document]
scottaohara: yeh.  but they're only right in that it's not document*s*.  SPAs are still very much still a web document... it just keeps getting rewritten instead of allowing navigation between different documents.

It's really just a silly thing to argue about imo. :)

checklist.html
ericwbailey: Is there another SC it would fail?

checklist.html
ericwbailey: I still think it's good UX to do, especially to make developers aware that it's A) something they can do and B) something that's useful. What's a better way to phrase this?

checklist.html
ericwbailey: Done.

checklist.html
ericwbailey: Would a followup sentence along the lines of, "Make sure you also don't interfere with the <kbd>Space</kbd> key's ability to scroll the page/view when not focusing on a form control." work for you?

checklist.html
scottaohara: >The text of the <code>title</code> element is typically one of the first announcements made by a screen reader when a new page loads. Adding concise feedback about the presence of errors or warnings, after a form submission has taken place, can help alert people that there are form fields that need attention.  Note that dynamically updating the text of a <code>title</code> element will not be announced to screen readers. SPAs will need to employ other methods to ensure such updates are announced by screen readers.

checklist.html
scottaohara: mmm.  maybe.  I think using the term "global" is still misleading though.  The <kbd>Space</kbd> key should really only work in the context of the media group.

checklist.html
scottaohara: i don't think so.  More of a best practice.  Especially regarding external links.

checklist.html
ericwbailey: Let's remove the reference to a global function, then:

>Provide the ability to pause any media control. If the device has a keyboard, ensure that pressing the <kbd>Space</kbd> key can pause playback. Make sure you also don't interfere with the <kbd>Space</kbd> key's ability to scroll the page/view when not focusing on a form control.

Sound good?

checklist.html
ericwbailey: So I think our options here are to:

1. Leave it be and have the SC be a little bit of a white lie, because it is an important thing to think about.
1. Remove it, as it's not technically a failure, which is not in keeping with our other checklist items.
1. Add a note to the end about what you mentioned, which I think kind of is like the second option.

I'm personally leaning towards removal. I want this checklist to be airtight.

checklist.html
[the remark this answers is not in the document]
ericwbailey: Just broke `autofocus` out into it's own item immediately after this one, if you'd like to check it.

I can't think of an article off the top of my head about roving tabindexes, does TPG have one we could use?

checklist.html
scottaohara: a lot better.  The only thing that is stands out to me now is calling it a "media control".   Swap out "control" with "element" and i think this is good to go.

checklist.html
scottaohara: I think this could stay in if you remove mentioning external links. Opening in a new window you could link over to http://www.w3.org/TR/WCAG20-TECHS/G201.html as a reasoning/ technique

scottaohara: I think this could stay in if you remove mentioning external links. Opening in a new window you could link over to http://www.w3.org/TR/WCAG20-TECHS/G201.html as a reasoning/ technique

ericwbailey: Done.

ericwbailey:
Merged. If there's further feedback let's address it on an Issue by Issue basis.

svinkle:
Awesome work @ericwbailey and @scottaohara! 👏 

I feel like we'll be able to improve upon this over time in a variety of ways. Perhaps by providing "further reading" links for more info and code snippets within the `details` content sections.

For example, to help with the Viewport list item, we could show what to do and what _not_ do to:

> 🛑 Don't do this:
```html
<meta name="viewport" content="initial-scale=1,minimum-scale=1,maximum-scale=1,user-scalable=no">
```

> ✅ Do this:
```html
<meta name="viewport" content="width=device-width, initial-scale=1">
```

The content that we have now is great, but doesn't really explain what _exactly_ to look for or how to solve the issue. I'm just thinking if someone new to all of this reads the list, it _may_ lead to more questions.

scottaohara:
I think that's a good idea for some of these that lend themselves to simple code dos/don'ts.

Was also just thinking this might provide us a bit of a road map on additional informative posts to add to the site as well.

ericwbailey:
I love those ideas! And thank you for the great new checklist content, @svinkle! We couldn't have done it without you 😀

I started an issue to capture/discuss it: https://github.com/a11yproject/a11yproject.com/issues/788

PR #782, opened 2019-04-30
Added how-to enable keyboard navigation in Chrome

mrazzari:
Same as Safari.


ericwbailey:
Looks good to me! @scottaohara?

PR #787, opened 2019-05-11
Add redirect functionality

ericwbailey:
This PR adds a gem to allow for redirects, as post filenames should reflect the title in their frontmatter.

I have also updated the title on an article to [adhere to our content styleguide's rules about the word "user"](https://github.com/a11yproject/a11yproject.com/blob/gh-pages/CONTENT_STYLE_GUIDE.md#user-or-person).

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -4,4 +4,5 @@ gem 'github-pages', group: :jekyll_plugins
 group :development, :test do
   gem 'rake'
   gem 'html-proofer'
+  gem 'jekyll-redirect-from'
 end
--- a/Gemfile
+++ b/Gemfile
@@ -4,5 +4,4 @@ gem 'github-pages', group: :jekyll_plugins
 group :development, :test do
   gem 'rake'
   gem 'html-proofer'
-  gem 'jekyll-redirect-from'
 end
--- a/_config.yml
+++ b/_config.yml
@@ -31,8 +31,9 @@ google_analytics:
 #   id: disqusid
 
 plugins:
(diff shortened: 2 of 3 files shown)


Gemfile
```diff
@@ -4,4 +4,5 @@ gem 'github-pages', group: :jekyll_plugins
 group :development, :test do
   gem 'rake'
   gem 'html-proofer'
+  gem 'jekyll-redirect-from'
```
nschonni: I believe this is already bundled with the `github-pages` gem

ericwbailey: I'm going off of this documentation: https://help.github.com/en/articles/redirects-on-github-pages

nschonni: Yup, I just don't think you need to add it to the gemfile, the rest looks good
> Make sure you're using the latest version of the GitHub Pages Gem, which contains these dependencies and plugins.
This also shouldn't hurt though unless the versions between the GitHub Pages gem and the other gem get out of sync. Don't think that should happen because there is no explicit version for either

ericwbailey:
Removed the redirect gem, as we already have the github-pages gem. I'll test it to confirm the redirect works.

_config.yml
```diff
@@ -32,7 +32,6 @@ google_analytics:
 
 plugins:
   - jekyll-feed
-  - jekyll-redirect-from
```
nschonni: Think this needs to stay

scottaohara: I added it back in via a separate commit.  didn't seem to correct the issue though.

PR #790, opened 2019-05-15
Add title attribute checklist item

ericwbailey:
This PR adds a checklist item to providing guidance about proper use of the `title` attribute.

### DIFF
--- a/checklist.html
+++ b/checklist.html
@@ -215,6 +215,19 @@ permalink: checklist.html
         </p>
       </div>
     </details>
+    <details id="remove-session-timeouts">
+      <summary>
+        Remove <code>title</code> attribute tooltips.
+      </summary>
+      <div class="checklist-details">
+        <p class="checklist-criterion">
+          <a href="https://www.w3.org/TR/UNDERSTANDING-WCAG20/ensure-compat-rsv.html">4.1.2 Name, Role, Value</a>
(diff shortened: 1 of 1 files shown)


checklist.html
```diff
@@ -215,6 +215,19 @@ <h2 id="label-global-code">Global code</h2>
...
+    <details id="remove-session-timeouts">
+      <summary>
+        Remove <code>title</code> attribute tooltips.
+      </summary>
+      <div class="checklist-details">
+        <p class="checklist-criterion">
+          <a href="https://www.w3.org/TR/UNDERSTANDING-WCAG20/ensure-compat-rsv.html">4.1.2 Name, Role, Value</a>
+        </p>
+        <p class="checklist-description">
+          The <code>title</code> attribute has <a href="https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/title#Accessibility_concerns">numerous issues</a>, and should not be used to create tooltips. Acceptable uses for <code>title</code> include labeling an <code>iframe</code> element or controls in a data table.
```
scottaohara: wordsmithing for consideration

>...should not be used if the information provided is important for all people to access.

>... or controls in a data table where the accessible name may be visually inferred from table headers, and would not need to be visually consistent in all table cells.

PR #794, opened 2019-05-17
[understanding-vestibular-disorders.md] Add note on 'prefers-reduced-motion' media query.

### LINKED ISSUE #792: Article: "A primer to vestibular disorders" should mention `prefers-reduced-motion` MQ
IMO, the section on [What should you consider](https://github.com/a11yproject/a11yproject.com/blob/gh-pages/_posts/2013-05-15-understanding-vestibular-disorders.md#what-should-you-consider) should mention the `prefers-reduced-motion` media-query:

- https://developers.google.com/web/updates/2019/03/prefers-reduced-motion
- https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-motion

I suppose it could be just an addition, however, I don't find the current advice of: only display animations after user-interaction (such as button click) a realistic goal for developers. This media query should (in [browsers that support it](https://caniuse.com/#search=prefers-reduced-motion)) negate the need for such on-site opt-ins.

Malvoz:
Resolves #792.

Feedback is welcome, of course.

### DIFF
--- a/_posts/2013-05-15-understanding-vestibular-disorders.md
+++ b/_posts/2013-05-15-understanding-vestibular-disorders.md
@@ -51,7 +51,7 @@ img {
 }
 
 button {
-  transition: rotate 1s;
+  transition: transform 1s;
 }
 
 button:focus,
--- a/_posts/2013-05-15-understanding-vestibular-disorders.md
+++ b/_posts/2013-05-15-understanding-vestibular-disorders.md
@@ -50,9 +50,13 @@ img {
   }
 }
 
+button {
+  transition: rotate 1s;
+}
+
 button:focus,
 button:hover { 
-  transform: scale(1.5); 
+  transform: rotate(360deg);
 }
 
 @media (prefers-reduced-motion: reduce) {
--- a/_posts/2013-05-15-understanding-vestibular-disorders.md
+++ b/_posts/2013-05-15-understanding-vestibular-disorders.md
@@ -33,21 +33,33 @@ The cause may be from illness, injury, or a genetic condition, but anyone can su
 
 Don't make animations, sliders, videos, or rapid movement start automatically. Give an indicator of what movement will happen on the site when someone takes action. Allow the option to turn off any animation and movement at any point in the process.
 
-Also, with the CSS [`prefers-reduced-motion`](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-motion) media query you can write conditional CSS animations and transitions based on the user's preference. For example, you may want to enable [smooth scrolling](https://developer.mozilla.org/en-US/docs/Web/CSS/scroll-behavior), but disable it and all other potential animations and transitions for users who explicitly prefers reduced motion:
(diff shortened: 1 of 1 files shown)


_posts/2013-05-15-understanding-vestibular-disorders.md
```diff
@@ -31,4 +31,28 @@ The cause may be from illness, injury, or a genetic condition, but anyone can su
 
 ## What should you consider?
 
-Don't make animations, sliders, or rapid movement start automatically. Give an indicator of what movement will happen on the site when someone takes action. Allow the option to turn off any animation and movement at any point in the process.
+Don't make animations, sliders, videos, or rapid movement start automatically. Give an indicator of what movement will happen on the site when someone takes action. Allow the option to turn off any animation and movement at any point in the process.
+
+Also, with the CSS [`prefers-reduced-motion`](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-motion) media query you can write conditional CSS animations and transitions based on the user's preference. For example, you may want to enable [smooth scrolling](https://developer.mozilla.org/en-US/docs/Web/CSS/scroll-behavior), but disable it and all other potential animations and transitions for users who explicitly prefers reduced motion:
```
ericwbailey: I like this addition a ton, but could we use an example other than smooth scrolling? 

It's not a very well-loved feature accessibility-wise, so I'd like to have the article not even suggest the hint that it's a property to use.

Malvoz: @ericwbailey 

I've included both a transition and an animation, and removed `scroll-behavior`. I also clarified that it is in the browser settings that the preference is exposed from.

However, I don't think it's the right approach to not expose developers to certain CSS properties because they may inherently be considered worse in terms of accessibility than the absence of that property. I'd rather create awareness around the potential accessibility issue(s) of that property/value pair and show how to (in this case) use a preference-based media query to restore the default and least annoying behavior.

But the point of the update to this article is to at least mention `prefers-reduced-motion`, which we're doing here :)

_posts/2013-05-15-understanding-vestibular-disorders.md
```diff
@@ -50,9 +50,13 @@ img {
   }
 }
 
+button {
+  transition: rotate 1s;
```
scottaohara: this would need to be `transform`

Malvoz: Ah, yes.

ericwbailey:
This is great! Thank you for the edit, @Malvoz!

Malvoz:
@ericwbailey @scottaohara

I think it'd be wise to add `animation-iteration-count: 1 !important` per https://github.com/mozdevs/cssremedy/issues/11#issuecomment-462867630.

ericwbailey:
Sure! I could make a quick tweak, with your permission. Otherwise, happy to merge a quick PR.

Malvoz:
Go ahead! :)

ericwbailey:
All set, thank you! https://github.com/a11yproject/a11yproject.com/commit/efb098a1b33e6ccff78ec05932e5807e975292ca

PR #796, opened 2019-05-22
an attempt to redirect

scottaohara:
well @ericwbailey, i say we give this a try at some point.

I will need to rename this branch as "master" so we can do a quick switch between the gh-pages and master branches in case this doesn't work as it did in the test repo, since GH doesn't allow for other branches to become the gh-pages source.


ericwbailey:
I'm good to give this a shot. Let's plan for a day and time where the ol' switcheroo will affect the smallest amount of users if things go poorly.

PR #797, opened 2019-05-24
Describe use case for table differently

hidde:
I worry “looks like a table” may be too broad. My layout also looks like a table, but using `table` would not be right. 

So my proposal would be to add something that's more about displaying data. What about: ‘display data in columns and rows’?


ericwbailey:
Love it. Thank you!

hidde:
🎉

PR #798, opened 2019-05-25
Closes #503 - Update Article 'Never Remove CSS Outlines'

### LINKED ISSUE #503: inaccuracy in article never-remove-css-outlines
While `Style the element itself. You can remove the outline as long as you style the focused element differently` is a valid approach to dealing with removal of outlines, the examples shouldn't use color change as an example. This directly contradicts WCAG2.0 1.4.1 and in some cases is a specific failure criteria (links and interactive components).

https://www.w3.org/TR/WCAG20/#visual-audio-contrast-without-color

> *1.4.1* Use of Color: Color is not used as the only visual means of conveying information, indicating an action, prompting a response, or distinguishing a visual element. (Level A)

https://www.w3.org/TR/WCAG20-TECHS/F73.html

> The objective of this failure is to avoid situations in which people who cannot perceive color differences cannot identify links (when people with color vision can identify links). Link underlines or some other non-color visual distinction are required (when the links are discernible to those with color vision).

> While some links may be visually evident from page design and context, such as navigational links, links within text are often visually understood only from their own display attributes. Removing the underline and leaving only the color difference for such links would be a failure because there would be no other visual indication (besides color) that it is a link.

I'll draft a PR tonight with alternative examples and specific language about avoiding color-only changes if there's no issues.

nchan0154:
#503 - Small update to make note of the fact that you cannot rely on color alone to display focus state. Also, I wasn't sure if I should be adding myself to authors.yml as this is a one liner. 😂

### DIFF
--- a/_posts/2013-01-25-never-remove-css-outlines.md
+++ b/_posts/2013-01-25-never-remove-css-outlines.md
@@ -30,7 +30,7 @@ If you do not like the default focus outline that is displayed when a user click
 
 1. Style the outline. Webkit browsers have a more prominent glow so you could try [styling it](https://developer.mozilla.org/en-US/docs/CSS/outline) to make it less obtrusive. Consider the use of `a:focus { outline: thin dotted; }` to normalize the look of the outline across browsers.
 
-2. Style the element itself. You can remove the outline as long as you style the focused element differently (using a combination of `color`, `background-color`, `border` or `text-decoration: underline` for example). When using this approach, [avoid using color alone](https://www.w3.org/TR/WCAG20/#visual-audio-contrast-without-color) as the only visual means of determining focus, as colorblind people may not be able to distinguish the focus state from the inactive state.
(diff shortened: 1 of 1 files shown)


_posts/2013-01-25-never-remove-css-outlines.md
```diff
@@ -30,7 +30,7 @@ If you do not like the default focus outline that is displayed when a user click
 
 1. Style the outline. Webkit browsers have a more prominent glow so you could try [styling it](https://developer.mozilla.org/en-US/docs/CSS/outline) to make it less obtrusive. Consider the use of `a:focus { outline: thin dotted; }` to normalize the look of the outline across browsers.
 
-2. Style the element itself. You can remove the outline as long as you style the focused element differently (using `color`, `background-color`, `border` or `text-decoration: underline` for example).
+2. Style the element itself. You can remove the outline as long as you style the focused element differently (using a combination of `color`, `background-color`, `border` or `text-decoration: underline` for example). When using this approach, [avoid using color alone](https://www.w3.org/TR/WCAG20/#visual-audio-contrast-without-color) as the only visual means of determining focus, as colorblind people may not be able to distinguish the focus state from the inactive state.
```
scottaohara: let's point the link to wcag 2.1: https://www.w3.org/TR/WCAG21/#use-of-color

scottaohara:
@ericwbailey, this looks good to me.  any other requests before merge?

ericwbailey:
Nope! This is great, thank you.

PR #807, opened 2019-06-03
Closes #804 - TechDebt: Clean-up Footer Styling

### LINKED ISSUE #804: TechDebt: Clean-up Footer Styling
This ticket involves 4, very small, code-level issues I found in the a11yproject.com site footer that I feel can be fixed quickly, carries near zero risk for the site, and reduces tech debt.

After completing the work detailed below, a typical site user should not notice a thing while a developer/contributor might appreciate these small issues being cleaned-up.

## Summary
1. Remove inline `margin-bottom: 0;` CSS from footer HTML
2. Remove unnecessary `.row` class from footer HTML
3. Swap ID for class in footer HTML, CSS, and JS
4. Remove duplicate styles from `master.scss`

### 1. Remove inline `margin-bottom: 0;` CSS from footer HTML
  Add a new class to `_sass/_footer.scss` for setting `margin-bottom: 0;` and add this new class as needed in `_layouts/default.html`.

### 2. Remove unnecessary `.row` class from footer HTML
  Perhaps an oversight however, the `<ul>` of contributors in the footer has the class `row` applied. Since this `<ul>` is already the child of a `<div>` with the class `row` applied, re-applying it to the  `<ul>` is unnecessary. Removing this class just makes things cleaner and more clear. I believe this class was never needed in the first place. 

### 3. Swap ID for class in footer HTML, CSS, and JS
Working with the same chunk of HTML mentioned in number 2.
...and I swap the reference to ID`#continubutors-list` to class `.continubutors-list` in the listed JS and SCSS file. This now matches how we're referencing other, similar areas of the footer in our JS and SCSS files. Plus classes are nice and flexible, IDs are not.

### 4. Remove duplicate styles from `master.scss`
If we look in `css/master.scss` and then in `_sass/_notification.scss`, we'll notice duplicated style declarations. Removing these styles from `css/master.scss` not only removes the dupe but gets rid of poor code style (style declarations should not appear in our main SASS stylesheet which is meant only for importing other styles).

## Contributor notes
I've already done the work above in my local copy of the A11y site but wanted to follow our own process for [fixing things](https://github.com/a11yproject/a11yproject.com/blob/gh-pages/CONTRIBUTING.md#fixing-things) (plus I'm just so bored at the moment). I do however, still need to fully test my work in more platform/browser/device combinations.

Otherwise, feedback welcome!

webuxr:
This PR should be MUCH cleaner than my previous. Apologies for the added noise.

### DIFF
--- a/_layouts/default.html
+++ b/_layouts/default.html
@@ -92,10 +92,9 @@
   <body>
     <svg style="display: none;">
       <symbol id="icon-github" viewBox="0 0 32 32" fill="currentColor">
-        <path d="M0 18 C0 12 3 10 3 9 C2.5 7 2.5 4 3 3 C6 3 9 5 10 6 C12 5 14 5 16 5 C18 5 20 5 22 6 C23 5 26 3 29 3 C29.5 4 29.5 7 29 9 C29 10 32 12 32 18 C32 25 30 30 16 30 C2 30 0 25 0 18 M3 20 C3 24 4 28 16 28 C28 28 29 24 29 20 C29 16 28 14 16 14 C4 14 3 16 3 20 M8 21 A1.5 2.5 0 0 0 13 21 A1.5 2.5 0 0 0 8 21 M24 21 A1.5 2.5 0 0 0 19 21 A1.5 2.5 0 0 0 24 21 z "></path>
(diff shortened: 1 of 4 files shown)


ericwbailey:
Agreed. Thank you, @webuxr!

PR #811, opened 2019-06-20
Fix "Our Purpose" link in footer

kimadactyl:
Link is redirecting to the wrong place when not at the site root.


ericwbailey:
Thank you!

PR #818, opened 2019-07-05
Gulp error handling

jeryj:
I added gulp-plumber to the dev dependencies and integrated it into `gulpfile.js`. Now whenever you do something like leave a comma off a `.scss` file it will log the error to the console _without_ crashing gulp watch.


ericwbailey:
Thank you!

PR #842, opened 2019-09-13
Fixes #837

suzannehillman:
As per the discussion in #837, I changed the contribute link to point to README.md instead of the base github page, and got rid of extra spaces around the class attribute.

PR #843, opened 2019-09-18
Contributing updates

ericwbailey:
This PR addresses issues with contributing to the project, as described by @suzannehillman in https://github.com/a11yproject/a11yproject.com/pull/842. It:

- Includes information about Forking the project to be able to submit Pull Request.
- Updates the clone instructions to include `git clone`, as well as using `https:` instead of `git:`.
- Adds messaging about how to request a non-Markdown/GitHub article writing process.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -40,7 +40,7 @@ People working on more sophisticated aspects of the site (new features, bug fixe
 If you are looking for a place to get started learning about Git and contributing to an Open Source project, [Egghead.io offers a great introductory course](https://egghead.io/courses/how-to-contribute-to-an-open-source-project-on-github).
 
 ### Forking
-To submit updates and changes to The A11Y Project, you will first need to [Fork the repository](https://help.github.com/en/articles/fork-a-repo).
+To submit updates and changes to The A11Y Project, you will first need to [Fork the repository](https://help.github.com/en/articles/fork-a-repo). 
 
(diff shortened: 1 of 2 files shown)


suzannehillman:
Looks good to me! Includes info about forking and why to do it, and lets people clone their copy without an ssh key. I can see why you might want writers of content not have to go through git or use MarkDown, so that part makes sense. 

I'm not sure how much instruction you want to be including - for example, my GUI reminds me to stage my changes and add a change comment before committing them, so I don't have to remember the next steps after a change. For those using the command line, would it be worthwhile to remind them of the steps involved with getting to a pull request? Or would you rather use the links to git usage? 

I largely ask because it can be confusing to use the command line for git even if one is familiar with it, and in all probability you want to help new people - and people who haven't used git in a while - succeed. If you like, I can help by listing the steps that I took - that way you don't have to remember things that your fingers or shortcuts or macros do for you.

ericwbailey:
Thanks for your review!

We're intentionally drawing the line at providing information and support for digital accessibility, and not version control/Git. It's outlined in the [Contributing documentation](https://github.com/a11yproject/a11yproject.com/blob/gh-pages/CONTRIBUTING.md#git-and-github).

The reason for this is to help keep the project scoped and focused. There's plenty of great resources for learning version control out there—by linking out to some of those resources and drawing the line at this level of support for Git we ensure our site maintainers can stay focused on the site's mission.

PR #851, opened 2019-10-19
Added t12t event in Stockholm Sweden

duff92:
Accessibility event in Stockholm Sweden with 500 open spots.


ericwbailey:
Thank you, and best of luck with the event!

PR #860, opened 2019-11-16
Closes #859: macOS browser keyboard nav

### LINKED ISSUE #859: Article Update: Safari keyboard navigation
Proposal for 3 Updates:

1. Rename article. This article covers keyboard navigation for all browsers on MacOS, not just Safari. Suggested name: Browser Keyboard Navigation in MacOSX. (Alternatively: make a new article for each browser)
2. Screenshots are now out of date. As of Catalina, there is only one option: "Use keyboard navigation to move focus between controls". Suggestions: Either replace the screenshot, or add an up-to-date version alongside older one.
3. Firefox. I'm not sure if/when checking the "Use keyboard navigation..." in Settings > Keyboard shortcuts became insufficient, but firefox will not tab to links (firefox 70, Catalina). To fix this, add the [accessibility.tabfocus config item](https://developer.mozilla.org/en-US/docs/Mozilla/Preferences/Preference_reference/accessibility.tabfocus) and set it to 7 in firefox about:config. I am _guessing_ that Firefox accessibility.tabfocus is 2 (Give focus to all form elements except text fields) even when "Use keyboard navigation..." is set, so links still wouldn't receive focus. 

I am happy to update the article

erinoerin:
Changes:
# Renamed Article

- Updated article title
- Updated file name
- Added redirect in frontmatter from old article path

# Slight reorganization of article

Grouped all macOS setting related sections together (terminal instructions were after browser config instructions). Grouped all browser related sections together.

# Added Catalina Information

Added screenshot for Catalina settings. I left the old screenshot and instructions for those waiting to upgrade.

# Added Firefox setting information

Added instructions and screenshots for both enabling and disabling the firefox config option


ericwbailey:
This is excellent, thank you so much for doing it!

PR #867, opened 2019-11-27
Update resources.html

LemonAndroid:


ericwbailey:
@LemonAndroid Please give some context to this change, and why you're submitting it.

LemonAndroid:
It's not meant to be merged (at all).

This is a request to consider the lesson of trying to make a QR code usable across all disabilities.


The sole goal of this challange is to realize, that we are not very far yet regarding aria- attributes.


https://stackoverflow.com/questions/59066426/how-to-enhance-the-accessibility-of-qr-codes-on-the-web


Thanks for answering.

PR #885, opened 2020-03-03
Add resources and shortcuts in VO and NVDA articles.

### LINKED ISSUE #674: Article Audit: NVDA + VoiceOver articles
In reviewing the NVDA and VoiceOver articles for #669, there are some updates I think would be good to create a level of consistency between the two, as well as to provide a template to follow for future articles for JAWS and mobile screen readers.

The NVDA article (2015) is largely good as is, but adding in additional links to more in-depth usage articles/resources would be good here.

The VoiceOver article (2013) could use a listing of common key commands, like the NVDA article, and also links to more in-depth usage articles/resources.

bistaastha:
Fixes #674 

Added a table for shortcuts in VoiceOver article and added a couple of resource links in both NVDA and VO articles.


ericwbailey:
This looks great, thank you very much for doing it!

PR #947, opened 2020-07-15
Resolve JS and Style linting issues

esjay:
This fixes a few in-place JS linting issues, as well as re-enables style linting which was broken at one point per the [roadmap card](https://github.com/a11yproject/a11yproject.com/projects/1#card-16261896). 

Note that enabling style lint uncovers 184 style errors. If we would rather wait on this until the errors are resolved, I can remove that change to get the JS lint issues fixed.


ericwbailey:
Amazing, thank you!

Issue #969, opened 2020-07-16
Jerry Jones's website links incorrectly to Eric Bailey's website


jina:
1. Go to [https://www.a11yproject.com/team/](https://www.a11yproject.com/team/)
2. Scroll down to Jerry Jones and click "website"

### Expected result

Expected link to Jerry's website.

### Actual result

Was taken to Eric Bailey's website.

I don't know Jerry's actual website URL, otherwise I would have submitted a pull request.

ericwbailey:
Hi Jina! Thanks for submitting this. Closed in https://github.com/a11yproject/a11yproject.com/pull/970, thank you to @danielmatthew for the fix.

PR #970, opened 2020-07-16
Fix: Update URL to Jerry Jones' website

danielmatthew:
As raised by @jina in #969, this updates the URL to point in the right place


ericwbailey:
Amazing, thank you.

PR #973, opened 2020-07-16
Fix #936: Use version field from package.json in build process

### LINKED ISSUE #936: Use semantic versioning for cache busting
We are now using [semantic versioning](https://semver.org/) for the site. This provides us with an excellent opportunity to tie the site's versioning to situations where the number can be used for things like cachebusting our CSS.

Currently, the site uses a manual `yyyymmdd` format to accomplish this. It would be great if we could expose the semver number as a variable and use that instead. This way, updating the site's version also automatically updates caching, etc.

danielmatthew:
Quickly spiking out an approach to resolve #936 

It appends the version field to the minified styles and scripts. What else needs cache-busting?


danielmatthew:
Per the deploy preview, that same version will need to be made accessible inside 11ty…

danielmatthew:
Hooray: the version now available inside 11ty and used in `base.njk`'s calls to the stylesheet and scripts.

ericwbailey:
@danielmatthew This looks good from my end, and really nicely done. Are you comfortable with me merging?

danielmatthew:
@ericwbailey I keep umm-ing and aah-ing about whether the version be part of the filename, or kept as a query parameter. I don't know if one is considered better practice than the other.

ericwbailey:
For what it's worth, I've used query parameter in the past to good effect. By which I mean I'm happy to own the fix if it blows up 😅

danielmatthew:
Right, in that case (😉), let me amend the URL structure tomorrow AM (UK time) and let's roll with it!

ericwbailey:
@danielmatthew Could I trouble you to update the merge conflict? Sorry about that, we had to do some emergency shuffling things around since your last update.

danielmatthew:
@ericwbailey sorted!

ericwbailey:
Amazing. Thank you so much!

PR #984, opened 2020-07-17
Resolve Stylelint errors

esjay:
This resolves outstanding errors thrown by Stylelint, either by attempting to coerce usage in code to meet the rule, or to adjust our own linting rules to match existing clean and practical usage.


src/css/components/_c-404.scss
```diff
@@ -11,7 +11,7 @@
 	background-color: $color-white;
 	border-top: $border-thicker solid $color-pink-shade;
 	margin-top: 2rem;
-	padding: 1rem 2rem 2rem 2rem;
```
ericwbailey, line 14: Looks like `padding-left` is getting dropped. Is this intentional?

src/css/components/_c-404.scss
```diff
@@ -28,7 +28,7 @@
 	}
 
 	@include mappy-bp(lap-small) {
-		padding: 1rem 4rem 2rem 4rem;
```
ericwbailey, line 31: Same here for `padding-left`.

src/css/components/_c-a11y-summary.scss
```diff
@@ -11,7 +11,7 @@
 	padding: 1.5rem;
 
 	@include mappy-bp(palm-large) {
-		padding: 2rem 7rem 2.75rem 7rem;
```
ericwbailey, line 14: Also here.

src/css/components/_c-copy-email-button.scss
```diff
@@ -9,7 +9,7 @@
 	font-size: 0.75rem;
 	letter-spacing: $font-tracking-slight;
 	margin-left: 1ch;
-	padding: 0.2rem 0.4rem 0.1rem 0.4rem;
```
ericwbailey, line 12: And here!

src/css/layout/_l-primary-nav.scss
```diff
@@ -28,14 +28,13 @@
 	// Supports
 	@supports #{$supports-flex} {
 		display: flex;
-		flex-direction: column;
 		flex-flow: row wrap;
 		margin-top: 0;
 
 		// Breakpoints
 		@include mappy-bp(palm-small) {
-			flex-direction: row;
```
ericwbailey, line 37: Looks like this is getting moved out of alphabetical order.

src/css/utilities/_spacing.scss
```diff
@@ -27,65 +27,6 @@
 }
 
 
-.u-spacing-shortest {
```
ericwbailey, line 30: I think these are currently commented out, but will be kept if we need them later on. Any reason for deleting them?

ericwbailey:
Hey @esjay, just a nudge on this. Is it something you'd like help with?

esjay:
> Hey Wayne Elgin, just a nudge on this. Is it something you'd like help with?

I can carve some time out this weekend to resolve your outstanding comments.

src/css/components/_c-404.scss
esjay: So, this is "intentional" but maybe I'm not understanding. If you're asking if the browser knows what to do, in the 3 value variant of the syntax, `padding-left` and `padding-right` are both set to the middle value.

If you are saying, you want to explicitly specify redundant values, we need to change the stylelint [rule `shorthand-property-no-redundant-values`](https://stylelint.io/user-guide/rules/shorthand-property-no-redundant-values) to allow for it. 

Can you advise which direction you prefer (removing the stylelint rule to allow redundant values or leaving this change as is)?

src/css/components/_c-404.scss
esjay: same as [prior response](https://github.com/a11yproject/a11yproject.com/pull/984/files/a442075085a0173abbc13b41848ff2affdf3df4c#r464021536) to other review.

Can you advise which direction you prefer (removing the stylelint rule to allow redundant values or leaving this change as is)?

src/css/components/_c-a11y-summary.scss
esjay: Same as [prior explanation](https://github.com/a11yproject/a11yproject.com/pull/984/files/a442075085a0173abbc13b41848ff2affdf3df4c#r464021536) of other review.

Please advise on same direction.

src/css/components/_c-copy-email-button.scss
esjay: Same as [prior response](https://github.com/a11yproject/a11yproject.com/pull/984/files/a442075085a0173abbc13b41848ff2affdf3df4c#r464021536).

Please advise on direction for all and I will adjust accordingly.

src/css/layout/_l-primary-nav.scss
esjay: This is directly related to the [rule `declaration-block-no-shorthand-property-overrides`](https://stylelint.io/user-guide/rules/declaration-block-no-shorthand-property-overrides) configured in our stylelint that `flex-flow: initial` is a shorthand which, by its placement afterwards, was overriding the `flex-direction: row` that was previously set before. In effect, `flex-direction: row` may be having no effect here because of that override.

As `flex-flow` is [just a combination](https://developer.mozilla.org/en-US/docs/Web/CSS/flex-flow) of `flex-direction` and `flex-wrap`, I think a fitting solution might be to just change `flex-flow: initial` to `flex-wrap: initial`. Would that be preferred?

src/css/utilities/_spacing.scss
esjay: These were purely duplicated rules that [were not commented out](https://github.com/a11yproject/a11yproject.com/blame/a442075085a0173abbc13b41848ff2affdf3df4c/src/css/utilities/_spacing.scss), so stylelint was complaining about them. 
Would you prefer that I leave the lines in place but comment them out?

Issue #950, opened 2020-07-15
Safari CSS issues


ericwbailey:
Our CSS needs a little love when it comes to Safari. Currently there are layout bugs for:

- [x] The `details`/`summary` elements for the Table of Contents component.
- [x] Focus state on sponsors
- [x] Layout of homepage cards on mobile

More added here as they're discovered.

andrewborstein:
Does `claimed` mean someone has already claimed responsibility for submitting PRs to solve the above issues, like perhaps @jendowns? If it's still available I'd be happy to open a PR for the Safari focus state bug.

ericwbailey:
@andrewborstein Yep. We're using `Claimed` to indicate the issue has someone working on it for sorting/filtering purposes. That being said, I was going to be the one who tackled the focus state bug. Would you like to do it instead?

andrewborstein:
Yeah, I already have a good lead on how to fix it, so happy to take it!

ericwbailey:
Closing this, as all issues have been addressed. Thank you @jendowns and @andrewborstein 🙏

PR #992, opened 2020-07-18
Fix Safari summary flexbox bug in Checklist and TOC

jendowns:
Partially addresses #950

Hey there! I noticed the [Safari flexbox bug for `summary` elements](https://bugs.webkit.org/show_bug.cgi?id=190065) is affecting the checklist page and also the table of contents, where elements (icon + text) that are meant to be flexed side-by-side are stacked:

Checklist | TOC
--- | ---
Screen Shot 2020-07-18 at 5 01 08 PM|Screen Shot 2020-07-18 at 5 01 17 PM

My understanding of this bug is that the `summary` element cannot be a flex container in Safari. To get around this, I have added a child `span` to the `summary` elements, and I have moved the flexbox styles to that `span`.

I called the `span` "summary-inner" but please let me know if you'd like for me to rename it. Thank you!


ericwbailey:
@jendowns Looks like there's a conflic in `toc.njk`. Do you want to address it, or shall I?

jendowns:
Thank you @ericwbailey! 🎉  I went ahead and resolved that conflict in my latest commit. Could you take another look when you get a chance?

ericwbailey:
Perfecto. Merged!

PR #994, opened 2020-07-18
Fix og:image tag URL

joelrodelo:
This PR address the `og:image` tag URL. It should be an absolute URL to work properly.


ericwbailey:
Great catch, thank you.

PR #1034, opened 2020-08-02
Close #1013: Add permalink to checklist items

### LINKED ISSUE #1013: Feature: Add direct permalinks for each checklist item
## Background
I often link to https://www.a11yproject.com/checklist when I want to share concise and easily digestable documentation about accessibility with my teammates. But I would love if there were a way to easily share a link directly to a specific checklist item, instead of the entire list or a [section of the list](https://www.a11yproject.com/checklist/#headings).

Currently my workaround is to grab the unique `id` value from the `<summary>` element in the DOM and manually append it to the URL. For example, https://www.a11yproject.com/checklist#dont-skip-heading-levels will link directly to that checklist item. 

From A11y Project:

Screenshot of a specific checklist item scrolled to the top of the viewport at a11yproject.com

## Potential implentation

One related technique that could be a source of inspiration is linkable headings.

### Heading text wrapped in a link

From A11y Style Guide:

Screensot of linkable heading at a11y-style-guide.com

Underlying markup (simplified):
```html
<h2>
  <a href="#kssref-structure-lists">
    Lists
  </a>
</h2>
```

Example link: https://a11y-style-guide.com/style-guide/section-structure.html#kssref-structure-lists

### Heading with icon link appended

From Andy Bell's blog:

Screen Shot 2020-07-22 at 6 29 20 AM

Underlying markup (simplified):
```html
<h2 id="heading-size-scales-ftw">
  Size scales FTW
  <a href="#heading-size-scales-ftw">
    <span class="visually-hidden"> permalink</span>
    <svg aria-hidden="true" focusable="false"><path d="M9.199 ..."></path></svg>
  </a>
</h2>
```

Example link: https://hankchizljaw.com/wrote/fluid-scale-and-tokens:-a-match-made-in-heaven/#heading-size-scales-ftw

### Other implementation thoughts

If the anchor link approach is appealing, I think it might need to live outside of the `<summary>` element, so it doesn't interfere with the normal open/close behavior. Or maybe there's a simple progressively enhanced JS solution that would work here, too, where opening a checklist item appends the `#unique-id` to the URL and visiting said URL can automatically open the item details if JS is enabled.

***

Is there any interest in adding a feature like this? If so, I'd love to discuss potential implementation ideas and would also be happy to take the lead on a pull request! 

BTW, love [the redesign](https://www.a11yproject.com/announcements/2020-07-15-2020-redesign) 🙌

andrewborstein:
Closes #1013

#### Permalink Example

Visitors on [the Checklist page](https://www.a11yproject.com/checklist/) can now 
1. Expose the URL to an individual checklist item by pressing its associated "Permalink" anchor element
2. Visit a checklist item permalink URL like `/checklist/#validate-your-html`, which scrolls the item into view and (when JS is enabled) opens its details

a11yproject.com checklist item opened and featuring new permalink
 
#### Notes and Questions
- I'm open to other UI suggestions! I wanted to aim for something simple, conventional, and matching existing styles. But I'm unsure if this link could benefit from some other visual distinction, particularly from the checklist item citation link.
- Is "Permalink" too jargony? I also considered copy like "Share Link" or "Shareable URL". 
- Does associating the link to its checklist item via `aria-describedby` sufficiently meet [2.4.4 Link Purpose (In Context)](https://www.w3.org/TR/WCAG20/#navigation-mechanisms-refs)? Including the entire title in the visible link text seemed liked overkill and rendering visually hidden text seemed unnecessary when the text already exists and could be linked using this technique. 
- I noticed a bug in Chrome where continually refreshing the page with a permalink URL in place continues scrolling further and further down the page instead of reloading the same scroll position. The issue doesn't occur when visiting the URL in new tabs or windows, or in Firefox or Safari. [It might be related to this bug 🤷 ](https://support.google.com/chrome/thread/11993079?hl=en), but doesn't seem like a critical user flow that needs to be supported above and beyond the first page visit always working.
- I wanted to extract `checklist.title | slugify` to a local variable but couldn't figure out how. I don't have a ton of Nunjucks / 11ty experience, so insight there is appreciated!

### DIFF
--- a/src/_includes/checklist.njk
+++ b/src/_includes/checklist.njk
@@ -1,4 +1,4 @@
-<details class="c-checklist">
+<details class="c-checklist" data-checklist-item-id="{{ checklist.title | slugify }}">
 	<summary id="{{ checklist.title | slugify }}" class="c-checklist__summary">
 		<span class="c-checklist__summary-inner">
 			<svg class="c-checklist__icon" aria-hidden="true" focusable="false" height="30" width="30" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 30 30"><path d="M9.07,29.55l-2.15-2.1L19,15,6.92,2.55,9.07.45,23.23,15Z"/></svg>
--- a/src/_includes/checklist.njk
+++ b/src/_includes/checklist.njk
@@ -12,6 +12,9 @@
 		{{ checklist.description | safe }}
 	</p>
 	<p class="c-checklist__share-link">
-		<a aria-describedby="{{ checklist.title | slugify }}" class="c-checklist__link" href="#{{ checklist.title | slugify }}">Share Link</a>
+		<a class="c-checklist__link" href="#{{ checklist.title | slugify }}">
+			Share Link
+			<span class="u-hide-visually">to checklist item: {{ checklist.title | safe }}</span>
+		</a>
 	</p>
 </details>
--- a/src/_includes/checklist.njk
+++ b/src/_includes/checklist.njk
@@ -1,20 +1,26 @@
-<details class="c-checklist" data-checklist-item-id="{{ checklist.title | slugify }}">
-	<summary id="{{ checklist.title | slugify }}" class="c-checklist__summary">
+{% set slugifiedTitle = checklist.title | slugify %}
+{% set description = checklist.description | safe %}
+{% set title = checklist.title | safe %}
+{% set url = checklist.url %}
+{% set wcag = checklist.wcag %}
+
+<details class="c-checklist" data-checklist-item-id="{{ slugifiedTitle }}">
+	<summary id="{{ slugifiedTitle }}" class="c-checklist__summary">
 		<span class="c-checklist__summary-inner">
 			<svg class="c-checklist__icon" aria-hidden="true" focusable="false" height="30" width="30" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 30 30"><path d="M9.07,29.55l-2.15-2.1L19,15,6.92,2.55,9.07.45,23.23,15Z"/></svg>
-			<span class="c-checklist__title">{{ checklist.title | safe }}</span>
+			<span class="c-checklist__title">{{ title }}</span>
 		</span>
 	</summary>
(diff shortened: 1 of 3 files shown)


ericwbailey:
Awesome, thank you for doing this. To answer your questions: 

> I'm open to other UI suggestions! I wanted to aim for something simple, conventional, and matching existing styles. But I'm unsure if this link could benefit from some other visual distinction, particularly from the checklist item citation link.

I like this approach. Simple, direct and to the point.

> Is "Permalink" too jargony? I also considered copy like "Share Link" or "Shareable URL".

I'm always of mixed opinions about the term permalink, but I think "Share Link" would probably be the way to go (especially if the checklist it's name ever updates 😅).

>Does associating the link to its checklist item via `aria-describedby` sufficiently meet [2.4.4 Link Purpose (In Context)](https://www.w3.org/TR/WCAG20/#navigation-mechanisms-refs)? Including the entire title in the visible link text seemed liked overkill and rendering visually hidden text seemed unnecessary when the text already exists and could be linked using this technique.

I'd actually like to use visually hidden text. I'm thinking both A) First rule of ARIA and B) I'm unsure how good the support would be for AT that generates a list of generated links. Mind updating it to use a visally hidden class?

>I wanted to extract `checklist.title | slugify` to a local variable but couldn't figure out how. I don't have a ton of Nunjucks / 11ty experience, so insight there is appreciated!

I believe slugify only works in Eleventy's loop through [Collections](https://www.11ty.dev/docs/collections/), but I can play around and see what I can do if you want.

andrewborstein:
Great points @ericwbailey! Thanks so much for the feedback. 

I changed "Permalink" to "Share Link" and removed the aria in favor of visually hidden text. I also found some documentation on [the `set` tag in Nunjucks](https://mozilla.github.io/nunjucks/templating.html#set), which I used to create local variables.

#### Updated "Share Link" Example

A11y Project checklist item opened and featuring new share link

When the example link is encountered by AT its accessible name should read, "Share Link to checklist item: Validate your HTML".

ericwbailey:
Love it. Thank you so much for doing this!

PR #1038, opened 2020-08-04
Ignore vendor lib JSHint errors

esjay:


ericwbailey:
Merged via a live pairing session.

PR #1039, opened 2020-08-04
Include Markdown files in purgeCSS options

esjay:


ericwbailey:
Reviewed in a live pairing session.

PR #1048, opened 2020-08-06
Remove markdownItAnchor

yatil:
Rationale: markdownItAnchor adds links with the hash symbol (#) to the headings of any page but those links are marked `aria-hidden="true"` which produces empty links that can be reached by keyboard. It is confusing for users of assistive technology. 

In an attempt to fix it one could add `tabindex="-1"` to those links, making them unavailable for keyboard users. But that would also be an issue.

I think the optimal solution is having a switch to turn those links on and adding an `aria-label` that says “Direct link to [this section] of the page”, with [this section] replaced by the name of the section.

Until a better solution exists I would remove the functionality as a whole.

### DIFF
--- a/.eleventy.js
+++ b/.eleventy.js
@@ -81,21 +81,13 @@ module.exports = function (eleventyConfig) {
 
   /* Markdown Plugins */
   let markdownIt = require("markdown-it");
-  let markdownItAnchor = require("markdown-it-anchor");
   let options = {
     html: true,
     breaks: true,
     linkify: true
   };
-  let opts = {
-    permalink: true,
-    permalinkClass: "direct-link",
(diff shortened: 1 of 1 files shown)


ericwbailey:
I don't disagree.

@yatil Any interest in helping me raising this as an issue on Eleventy's tracker? Seems like something we could work at to create a better default.

yatil:
@ericwbailey I think we need to bring it up in the markdownItAnchor repository, too: https://github.com/valeriangalliat/markdown-it-anchor

I don't think I have the bandwidth to do anything before the weekend, but feel free to start if you can and ping me. I’m happy to pinch in where I can.

PR #1052, opened 2020-08-07
Fix copy error within the animation checklist item

jeffstolz:
This commit removes a duplicate "not" within the animation title string.


ericwbailey:
Nice catch! Thank you for submitting a fix 🎉

PR #1065, opened 2020-08-16
Fix images missing file extensions on team page

jendowns:
Hey there, I noticed that 2 images were missing file extensions.

Let me know if I missed anything -- thank you!


ericwbailey:
Ah, great catch.

PR #1073, opened 2020-08-18
Add verb to touch target guidance

dengeist:
Most (all) of the other criteria in the checklist begin with verb phrases; this copyedit brings the touch target guidance in line with that trend.


ericwbailey:
Great catch! Thank you for submitting a fix.

PR #1082, opened 2020-08-23
Add custom form validation for Contact and Newsletter forms

### LINKED ISSUE #1057: Add better form validation for our newsletter and contact pages
Currently, we rely on the `required` attribute for our [newsletter signup](https://www.a11yproject.com/newsletter/) and [contact page](https://www.a11yproject.com/contact/) forms. While this does provide basic validation for user input, it is a less than ideal experience.

The public email address we were using was failing to deliver email, so we prioritized speed in getting a replacement up and running. We can do better, however. We need:

- [x] JavaScript validation logic for empty required input on form submission.
- [x] Clear and concise error messaging associated with each input via ARIA.
- [x] A summary area at the top of the form if more than one error is present, linking to each relevant input field.
- [x] A designed error state that does not rely on color alone.

jendowns:
Resolves #1057 

### Build
- Use `gulp-uglify-es` (this was a bit of a personal pref when writing these functions, to make things easier with template strings and so forth. Please let me know if you have any concerns about this vs `gulp-uglify`!)
- Do not purge styles with `c-form` class names (this is specifically to keep error message related styles from being removed)

### CSS
- Add styling for error wrapper, summary, and message (link) elements.

### Markup
- Use `novalidate` attribute on forms to prevent default HTML5 validation. This was necessary because otherwise, the HTML5 validation would take control after submitting invalid form fields -- it would generate its own tooltips & then move focus to the first invalid field. With `novalidate`, we can control focus after submit (and in this case, send focus to the first error message link).

### JS
- Add submit handler and functions to validate, add error messages, clean up, etc. Note that for the email validation, I kept the validation as close to the default HTML5 email validation as possible... it only requires one character + `@` + another character (i.e., like `a@a`). I think the only way to truly check an email address is to try to send an email to it, and this regex is more geared toward helping people catch common mistakes. 

That's a general summary of the changes. Please let me know what you think! I tried to add notes here and there, but let me know if they are too verbose, etc. Thank you!


src/contact.njk
```diff
@@ -17,13 +17,14 @@ templateClass: template-simple
 		</p>
 
 		<form
+			novalidate
```
ericwbailey, line 20: Nice.

src/css/components/_c-form.scss
```diff
@@ -101,6 +101,27 @@ $_c-input-margin-top: 1.5rem;
 }
 
 
+.c-form__error-wrapper {
+	border: 4px solid $color-red;
```
ericwbailey, line 105: Could we use a variable from [`_variables.borders.scss`](https://github.com/a11yproject/a11yproject.com/blob/main/src/css/logic/_variables.borders.scss) here? `$border-thin` seems like it might work.

src/css/components/_c-form.scss
```diff
@@ -101,6 +101,27 @@ $_c-input-margin-top: 1.5rem;
...
+	.c-form__error-message {
+		font-size: 0.75rem;
+
+		a {
+			@include link-states($color-black);
+		}
+	}
+}
+
+.c-form__input[aria-invalid]{
```
ericwbailey, line 121: I like how you've tied appearance to state here. My preferred way to do something like this.

src/css/components/_c-form.scss
```diff
@@ -101,6 +101,27 @@ $_c-input-margin-top: 1.5rem;
...
+	border: 2px solid $color-red;
```
ericwbailey, line 122: If we use border variables, we'd want to update `2px` to use `$border-thinner` here.

jendowns:
Thank you @ericwbailey! I just pushed a commit that updates the border variables. And I also added just a little more spacing between error list items.

Issue #948, opened 2020-07-15
Refactor profile photos


ericwbailey:
We use a circular photo treatment for photos of team members, authors, and industry spotlights. The underlying Sass is tied to the component the photo is placed on, and not the photo treatment itself.

This signals an opportunity to break the photo treatment out into its own component.

beckkramer:
I'd be interested in helping with this if this.

ericwbailey:
That would be wonderful, thank you!

ashutosh887:
Is this Issue resolved?
Or should I work on it !
@ericwbailey @beckkramer

beckkramer:
Hello! I should be able to resolve this this week, thank you for the nudge. And Eric, apologies for disappearing on this- the end of last year was difficult and I lost track of things.

ericwbailey:
@beckkramer No worries. Last year was tough on all of us, I hope you're doing okay.

PR #1099, opened 2020-09-08
Issue 948: Refactor profile photos

beckkramer:
Implements Issue #948 

Included in this PR:

- Adds template and Sass files related to circular avatar images
- Integration of these new style utility and template into existing components
- Moves from an `onClick` event on an image to an anchor element when a link is passed in


src/_includes/avatar.njk
```diff
@@ -0,0 +1,14 @@
+{% if avatar.link %}
+  <a
+    class="c-avatar {{ avatar.cssClass }}"
+    style="background-image: url({{ avatar.image }});"
+    href="{{ avatar.link }}"
+  >
+    <div class="u-hide-visually">{{ avatar.name }}'s avatar</div>
+  </a>
```
ericwbailey, line 8: I'd like to remove this and re-insitute the inline `onclick="location.href='{{ avatar.url }}"` JavaScript we're currently using.

The reasoning for this is the person name's is wrapped in an anchor element, and provides an accessible name for assistive technology. 

`{{ avatar.name }}'s avatar` does not provide much in the way of information, and adding it as an anchor introduces a second, redundant link. This can make traversing a list of links more annoying, or even cumbersome when severe motor control issues are present.

It's a little counterintuitive hack that makes for a better assistive technology experience 🙂

src/_includes/homepage/spotlight.njk
```diff
@@ -1,9 +1,13 @@
 <div class="c-homepage-card c-homepage-card--spotlight">
-	<img
-		class="c-homepage-card__spotlight-photo"
-		alt="Photo of {{ spotlight.name }}."
-		onclick="location.href='/spotlight/{{ spotlight.name | slugify | url }}'"
-		src="img/spotlight/{{ spotlight.image }}" />
+	{%
+		set avatar = {
+			cssClass: "c-homepage-card__spotlight__avatar",
```
ericwbailey, line 4: Can we reintroduce `c-homepage-card__spotlight-photo` as a class name? We try not to chain Elements within Elements with [BEM](http://getbem.com/introduction/).

src/_includes/homepage/spotlight.njk
```diff
@@ -1,9 +1,13 @@
...
+			image: "img/spotlight/" + spotlight.image,
+			link: "/spotlight/" + spotlight.name | slugify | url,
+			name: spotlight.name
+		}
+	%}
+	{% include "avatar.njk" %}
```
ericwbailey, line 10: Ooh, nice! I like what you're building towards here 😀

src/_includes/layouts/featured.njk
```diff
@@ -3,7 +3,15 @@ layout: layouts/base.njk
 templateClass: template-generic
 ---
 
-{% include "spotlight/hero.njk" %}
+<div class="l-spotlight__hero c-spotlight__hero">
+	{%
+		set avatar = {
+			cssClass: "c-spotlight__hero__image",
```
ericwbailey, line 9: Same concerns here for double element chaining in BEM.

src/css/utilities/_avatar-size.scss
```diff
@@ -0,0 +1,12 @@
...
+	"small": 6rem,
+	"medium": 8rem,
+  "large": 12rem,
+  "hero": 14rem,
+);
+
+@mixin avatar-size($size: "tiny") {
+	height: map-get($avatar_dimensions, $size);
+	width: map-get($avatar_dimensions, $size);
+}
```
ericwbailey, line 12: We can move this into `c-avatar`, since it is component-specific logic. Our preferred style is an underscore prefix for component-specific variables and mixins, as well as dashes instead of underscores for word spaces. So, it would be `$_avatar-dimensions` and `_avatar-size`.

I know this seems pedantic, but we try to keep component-specific logic inside the component its utilized in to help both with maintenance and to communicate it shouldn't be used elsewhere improperly. If we find that this logic is needed elsewhere it signals the need for a refactor.

ericwbailey (changes requested):
@beckkramer This is some really nice work, thank you. Some feedback around updating to match our preferred style, but I really like the approach you took.

src/_includes/avatar.njk
beckkramer: Gotcha- let me revert this to the past approach 👍 I think the case of the spotlight image being an `img` element with an alt text value vs the other instances being `div`s made me overthink things a bit here.

Appreciate the context for the original approach as well- that balancing act of reducing static while still giving enough context is always tricky, and I'm always interested to learn how others are approaching these kinds of situations.

src/css/utilities/_avatar-size.scss
beckkramer: Not pedantic at all, it's nice to have clarity in approach! That said, I'd like to chat about this a little more:

Is it expected to have mixins from one component's CSS being referenced in another component's? `avatar-size()` is referenced in several other Sass files, which was the reasoning behind peeling this piece of the CSS out into its own file: it felt more like a utility. It's the only part of the avatar component's CSS that isn't component-specific, but rather needs to be affected by its parent component.

I tend to interpret a mixin being in a component file to mean it's just for that file's use, which is how the rest of the codebase also seems to be set up; the only other non-third-party mixins I found in multiple files in the project were `hide()`, `link-states()`, `preserve-list-semantics()` and `heading-X()`, which aren't in component Sass files.  Maybe the avatar size mixin should be moved to the `src/css/logic/` folder? Or could be made more general as an image size mixin? Is there another aspect to this I'm missing?

Another option could be to just have the avatar component always be 100% height/width (perhaps with a max size in place as a fail-safe), and then it's up to the parent component to set the constraining value. The current sizes aren't providing a ton of consistancy as-is, so I'm not sure we'd lose much by eliminating the size map (it could just go back to being what was in the `_person.scss` file before). This might make things more self-contained while still having parents call the shots in sizing, which was the ultimate goal for all this. What do you think?

ericwbailey: The approach we're using is: 

Use an underscore for a mixin that's helpful to a single component. If more than one component could benefit from the mixin, promote it by creating a file for it in ['logic/`](https://github.com/a11yproject/a11yproject.com/tree/main/src/css/logic). So, based on your description, I think your call of moving it there is a good one.

As for the second option you propose, I think your initial call is a good one. I think this is more how to treat the mixin logic, and that's on me for not having good documentation. We have a little bit of it written down in [our styleguide](https://www.a11yproject.com/styleguide/), but it's something I'm slowly chipping away at when time permits.

beckkramer: Great, thanks for helping me understand! I should be able to get some work in on this this weekend for this and the other changes.

ericwbailey:
Closed for inactivity. Please comment if you are till interested in working on this and I'll reopen the issue.

src/css/utilities/_avatar-size.scss
beckkramer: Summary of the changes I ended up making here:

* Renamed the mixin file to `_mixins.avatar-size.scss`
* Moved it into the `/logic` folder

I _think_ this is what we landed on per the above conversation, but wanted to check in on it since so much time has passed. Happy to revise as needed!

beckkramer:
Hi Eric- I've just pushed up some changes based on your feedback, but let me know if you'd like any other adjustments when you can.

beckkramer:
@ericwbailey Just wanted to follow up on this when you have a moment.

Issue #989, opened 2020-07-18
Remove Gulp as a dependency


ericwbailey:
I love Gulp, but I love less external dependencies more. I initially used it for the redesign to get things up and running quickly.

The tasks we're using Gulp for are relatively straightforward, namely asset minification, concatenation, post-processing, and Browsersync. Because of this, there is an opportunity to do a lot of this via node scripts. Big-picture, I'd also like to lean more on Eleventy's capabilities to do the same—we're already using it for things like `addPassthroughCopy`.

To remove Gulp, we'd need to:

- [ ] Create node scripts for each Gulp task, and ensure feature parity.
- [ ] See what tasks can be used in Eleventy directly.
- [ ] Update `npm run a11yproject` to make sure it works.
- [ ] Update Netlify to reflect these changes.
- [ ] Remove references to Gulp in our documentation.
- [ ] Do a little happy dance.

ericwbailey:
@esjay You had mentioned wanting to tackle this previously. How does this sound to you?

esjay:
> @esjay You had mentioned wanting to tackle this previously. How does this sound to you?

@ericwbailey put me in coach!

ericwbailey:
Woo!

esjay:
Due to a dependency upon how eleventy processes incremental builds, we can get about 95% of the way towards this, but the build output will suffer due to eleventy's need to [rebuild the entire site](https://github.com/11ty/eleventy/issues/108) whenever a passthrough file copy operation occurs. Currently, this is on [eleventy's Incremental enhancement board/roadmap](https://github.com/orgs/11ty/projects/3). 

After discussing with @ericwbailey, we decided to pause existing work on this PR until something moves on the parallelized/incremental build front with eleventy.

SaptakS:
The updates to eleventy's [incremental usage](https://www.11ty.dev/docs/usage/incremental/) in v2.x might make this possible now?

ericwbailey:
Yes! Upgrading this site to use Eleventy 2 is on my list, and this should address that, exactly like what you said, @SaptakS.

@esjay: Is this something you'd be interested in pairing on?

PR #1141, opened 2020-10-27
Remove Gulp in Favor of NPM CLI

esjay:
In this PR, per Issue #989, we remove all traces of Gulp in favor of all NPM scripts for both development and build phases. We leverage new CLI-compliant NPM packages where necessary to provide cross-platform compatibility for non-unix-shell environments.


esjay:
putting this work on hold until #989 is addressed at the eleventy is in a better place with incremental build

nchan0154:
Hi @esjay do you have any further interest in working on this? I was looking into upgrading to 11ty v2 first before seeing if we could merge this, but I think due to the nature of how entwined these changes are, I think we need to do both of these changes at the same time. Let us know if you would still be interested in tackling this.

nchan0154:
Upgraded all the dependencies using the great base that Wayne has given us, and now the build does actually run well with Eleventy's incremental watching. I need to compare the site side by side and make sure everything still works, and double check that all the options to our new dependencies are valid/not deprecated.

nchan0154:
(As always, apologies for coming back to this so late.)
I've tested the site side by side and it looks good. However, I'm running into some issues with Windows compatibility. There's lots of references to things that don't work on Windows, eg.
- `svgo \"dist/img/**/*.svg\"` Glob syntax doesn't work in windows
- `mkdir -p` Mkdir does work on windows, but the `-p` flag doesn't. I've tested it with `npx mkdirp` and it does work.
- `mv` and `rm` don't work either

I'm wondering what you all feel the best path forwards is. Personally, I do like the idea of keeping it lean with `npm scripts`, but once we start using npx to pull in random packages for that cross system goodness, we end up creating our own little monster that may be harder than just using something like [Vite](https://www.11ty.dev/docs/server-vite/), which 11ty now supports. Open to thoughts and suggestions.

PR #1155, opened 2020-11-19
Puts toc text inside <span> tag

### LINKED ISSUE #1150: Code blocks in ToC
## Summary

On mobile, inline code in the table of contents does weird wrapping.

1. Use Firefox or Chrome on Android mobile,
2. Go to https://www.a11yproject.com/posts/2019-02-15-creating-valid-and-accessible-links/,
3. Scroll down to the _Table of Contents_,
4. Admire.

### Expected result

The text flows like normal text.

### Actual result

Items with a piece of code cause wrapping of all text before the code, then the code (which also wraps), then more wrapping.

Screenshot_20201112-212508

SaptakS:
Fixes #1150

I have put all the innerHTML content inside a `<span>` tag so that the flex can only be between the text and `:after` focus line. Also added `overflow-y: hidden` for code inside toc. This should solve all the problems.

I was trying to avoid putting a `<span>` inside an `<a>` tag, but maybe it's not too horrible? And feel like fixes the design bugs.

@ericwbailey @aardrian what do you think?

### DIFF
--- a/src/css/vendor/_v-toc.scss
+++ b/src/css/vendor/_v-toc.scss
@@ -167,10 +167,7 @@
 		code {
 			text-transform: none;
 			line-height: 1.1;
-
-			@supports not (gap: 0.5rem) {
-				margin: 0 0.5rem;
-			}
+			overflow-y: hidden;
 		}
 
 		li {
@@ -210,8 +207,6 @@
 		@supports #{$supports-flex} {
 			border-bottom: none;
 			display: flex;
-			flex-wrap: wrap;
-			gap: 0.5rem;
 
 			&::after {
 				content: "";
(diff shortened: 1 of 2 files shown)


aardrian:
I don't speak Sass (or whichever CSS compiler that is) but I made what I believe are the changes (`<span>` wrapper, removed `flex-wrap` and `gap` from `.v-toc__nav a`) in the dev tools in Chrome and the text is no longer broken. I did not do it in more browsers.

SaptakS:
> I don't speak Sass (or whichever CSS compiler that is) but I made what I believe are the changes (`<span>` wrapper, removed `flex-wrap` and `gap` from `.v-toc__nav a`) in the dev tools in Chrome and the text is no longer broken. I did not do it in more browsers.

That sounds right!

PR #1176, opened 2021-01-08
Typographical error fixes.

whitingx:
Replacing uses of ‘Accessibile’ with ‘Accessible’ on the Resources page.


ericwbailey (approved):
Nice catch, thank you!

PR #1187, opened 2021-02-04
Adding VoiceOver Testing resource

yakimvanzuijlen:
I wrote an article on how to use VoiceOver on macOS for the purpose of testing screen reader compatibility.

I see that there are already some similar articles out there already, so this might repeat a lot of the concepts. I think it could help people who think visually to understand how a screen reader (specifically VoiceOver) works. It's written from my own perspective as a sighted person about how I learned using VoiceOver and the things that I didn't fully understand at first.

### DIFF
--- a/src/_data/resources.json
+++ b/src/_data/resources.json
@@ -1359,6 +1359,12 @@
 			"description": "Answers to common questions.",
 			"additional": "WebAIM",
 			"url": "https://webaim.org/articles/screenreader_testing/"
+		},
+		{
+			"title": "How to start testing screen reader support using VoiceOver",
+			"description": "This article explains why you should be testing your content with a screen reader. It also gives a visual guide on how to use VoiceOver on macOS",
+			"additional": "Yakim van Zuijlen",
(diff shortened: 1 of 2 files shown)


ericwbailey:
Hi @yakimvanzuijlen, thanks for submitting this! 

I was actually reading it this morning, so it was a nice surprise to see it added to the project. Could I trouble you to also add it to our [Getting started with macOS VoiceOver](https://github.com/a11yproject/a11yproject.com/blob/main/src/posts/2013-01-14-getting-started-with-voiceover.md) post's `further_reading` metadata? I think it'd make a great addition there as well.

Also, mind if I add the post to our social media feed's queue?

yakimvanzuijlen:
Thanks for reviewing it Eric. I added it to the future reading section of the VO article. I'd definitely appreciate it if it will be shared on social media.

ericwbailey (approved):

ericwbailey:
Amazing. Thank you for submitting this!

Issue #1206, opened 2021-03-11
Typo in “Should I use an accessibility overlay?” post


adampage:
## Steps to reproduce

1. Open any browser.
2. Go to [Should I use an accessibility overlay?](https://www.a11yproject.com/posts/2021-03-08-should-i-use-an-accessibility-overlay/) at a11yproject.com.
3. Navigate to the [ Performance ](https://www.a11yproject.com/posts/2021-03-08-should-i-use-an-accessibility-overlay/#toc_Performance) section.

## Actual result

Observe typo (emphasis mine):
“...may cause the entire **_thing crash_**.”

## Discovered with

| Technology | Name | Version |
| --- | --- | --- |
| Device | MacBook Pro |15-inch, 2018 |
| OS | MacOS Big Sur | 11.2.2 | 
| Browser | Chrome | 88.0.4324.192 |
| Assistive technology | None ||

ericwbailey:
Closed via https://github.com/a11yproject/a11yproject.com/pull/1207. Thanks for spotting this!

PR #1207, opened 2021-03-11
Fix typo in “Should I use an accessibility overlay?” post

adampage:
This addresses issue #1206.  Thanks so much for this very helpful article!


ericwbailey (approved):

ericwbailey:
Nice catch, thank you!

Issue #1208, opened 2021-03-12
Inaccessibility: Low contrast .c-card__additional in Featured Resource Card


kelsS:
Please review our Contributing guidelines:
https://a11yproject.com/contributing-guidelines/

## Summary
When you right-click on the featured resource card link the text for the `.c-card__additional` paragraph changes to white over the green background of the card and has very low contrast.

> Brief description of the issue. Please identify specific steps.

> 1. Navigate to https://www.a11yproject.com/resources/ 
> 2. Hover over the main link in the featured resource card
> 3. Right-click on the main link in the featured resource card

### Expected result
Expect that the text color for `.c-card__additional` does not change to white.

### Actual result
The text turns to white and is then not readable.

![color-contrast-issue-a11yproject](https://user-images.githubusercontent.com/20908392/110964325-12f2e600-8321-11eb-8a65-c3900d5b4206.gif)

### Additional Information

> To help us try and reproduce the issue, please provide the following information:
> 1. What device were you using? (laptop, phone, tablet, etc.)
    Macbook Pro 16 inch

> 2. Operating System version.
   macOS Big Sur 11.2.3

> 3. Browser(s) and version(s).
   Firefox Developer Edition, latest version.

> 4. Assistive technology used (if applicable)

### Additional Information

Note anything else you think would be important to know that didn't fall into an above category.

ericwbailey:
Great catch, @kelsS! I'll get on this right now.

ericwbailey:
Closed via https://github.com/a11yproject/a11yproject.com/pull/1213. Thank you for reporting this!

kelsS:
Awesome! Thanks @ericwbailey :)

ericwbailey:
Reopened via https://github.com/a11yproject/a11yproject.com/pull/1213#issuecomment-801328990.

mxmason:
Hey, @ericwbailey, is this still an issue? I tabbed through the cards on all current macOS browsers and couldn't get the contrast issue to appear. I see that #1220 is claimed, but if you don't have time, I'm happy to give it a shot.

ericwbailey:
@mxmason Forgot to close this, but good to have it checked by a second person! Closed via https://github.com/a11yproject/a11yproject.com/pull/1239.

PR #1213, opened 2021-03-17
Update featured resource focus style

ericwbailey:
This PR addresses https://github.com/a11yproject/a11yproject.com/issues/1208. It updates the focus style of the featured resource to improve color contrast when in a focused state.


kelsS:
Hey @ericwbailey, just realized the fix for this might have caused a bug for the non-featured cards on the page or I just missed the issue last time I looked at the page. When I tab through the page the non-featured cards look like the following when in focus:

a11yproject-focus-bug

ericwbailey:
Shoot, confirmed on my end as well. I'll get on fixing the fix.

PR #1237, opened 2021-04-07
Updating Resources for Accessible360 for consistency.

jasonwebb:
On the [Resources page](https://www.a11yproject.com/resources/) I noticed two references to Accessible360 with two slightly different descriptions, and one of those referencing using the company's website URL as the company name. This PR is just to tweak those two references to be identical with some updated copy for the description!


ericwbailey (approved):
Nice catch, than you!

PR #1243, opened 2021-04-20
Block FLoC

ericwbailey:
This PR [opts The A11Y Project out of Google's FLoC trial](https://developer.chrome.com/blog/floc/#how-can-websites-opt-out-of-the-floc-computation).


SaptakS (approved):
Looks good to me. Merging.

PR #1244, opened 2021-04-22
Update tooling and package versions

mxmason:
This PR plays catchup with some software versions!

1. .nvmrc gets bumped to `14` because 14 is the active LTS, and we were 4 major releases behind.
2. Multiple packages get bumped to their latest versions. Of note, `@11ty/eleventy` is at `12.0.1`. This upgrade is a good idea because eleventy supports Node 10+ [beginning at major version 12](https://github.com/11ty/eleventy/releases/tag/v0.12.0).

**Bonus:** The upgraded version of `stylelint` caught an error it previously overlooked. This PR fixes that.


ericwbailey (approved):
This is a very welcome update, thank you.

PR #1246, opened 2021-04-22
Simplify date conversions

mxmason:
This PR removes the `moment` and `Luxon` dependencies, as Node now has better date powers by way of `Intl.dateTimeFormat`.

Handling the post dates with native code causes an error if date values are not provided. My change to `post.njk` adds a check to make sure `last_updated` exists.


ericwbailey (approved):
Yesssss.

PR #1247, opened 2021-04-23
Improve main script performance

mxmason:
This PR contains some overall improvements to `main.js`. Highlights:
- Less iteration, where possible
- More performant iteration by replacing `NodeList.forEach` with a `for...` loop
- Fewer nested levels of function calls

### DIFF
--- a/src/js/main.js
+++ b/src/js/main.js
@@ -72,7 +72,7 @@ function openLinkedCheckListItem() {
 }
 // Store checklist status ---------------------------------------------------
 function storeChecklistItem(checkboxId) {
-	localStorage.setcheckboxId(checkboxId, 'checked');
+	localStorage.setItem(checkboxId, 'checked');
 }
 
 function removeChecklistItem(item) {
@@ -88,7 +88,7 @@ function processChecklistClick(selector) {
 		}
 
 		if (target.checked) {
-			storeChecklistItem(checkboxId);
+			storeChecklistItem(target.id);
 		} else {
 			removeChecklistItem(target.id);
 		}
--- a/src/js/main.js
+++ b/src/js/main.js
@@ -58,13 +58,16 @@ if (navigator && navigator.clipboard && navigator.permissions) {
 	});
 }
 
-// When someone navigates directly to a checklist item using its "Share Link"
-// like a11yproject.com/checklist/#validate-your-html, the item with the
-// matching id attribute will be scrolled into view. Then, if JS is enabled,
-// this code will open its associated <details> element.
-function openLinkedCheckListItem(itemId) {
+/**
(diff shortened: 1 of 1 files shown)


ericwbailey:
@SaptakS JS isn't my strong suite, mind reviewing this?

src/js/main.js
```diff
@@ -50,64 +58,65 @@ if (navigator && navigator.clipboard && navigator.permissions) {
...
-function renderLocalStorageItems(items) {
-	items.forEach(function(element){
-		if (checkChecklistExists(element.id)) {
-			element.checked = true;
-		}
-	});
+function populateChecklistFromLocalStorage(checkboxSelector) {
+	var items = document.querySelectorAll(checkboxSelector);
+	var length = items.length;
+	for (var i = 0; i < length; ++i) {
```
SaptakS, line 105: Any particular reason for replacing the forEach with a for loop?

src/js/main.js
```diff
@@ -50,64 +58,65 @@ if (navigator && navigator.clipboard && navigator.permissions) {
...
+		var checkboxElement = items[i];
+		checkboxElement.checked = localStorage[checkboxElement.id] === 'checked';
```
SaptakS, line 107: I would prefer localStorage.getItem() instead. Though in this case it doesn't really make any difference, but I find it preferable to use the function defined.

src/js/main.js
```diff
@@ -50,64 +58,65 @@ if (navigator && navigator.clipboard && navigator.permissions) {
...
+ * (ex: a11yproject.com/checklist/#validate-your-html) the item with the
+ * corresponding id will scroll into view. Then, if JS is enabled,
+ * this function will open its associated <details> element.
+ */
+function openLinkedCheckListItem() {
+	var hash = window.location.hash.substr(1);
+	var checklistItem = document.querySelector(
+		"[data-checklist-item-id=" + hash + "]"
+	);
+
```
SaptakS, line 72: This is great improvement! Only in the comment section, I would change it to `If someone opens the `checklist` page using a checklist "Share link" (ex: a11yproject.com/checklist/#validate-your-html) the item with the corresponding id will scroll into view. Then, if JS is enabled, this function will open its associated <details> element.`

SaptakS:
The overall changes look great to me! I have left a few comments which are mostly like questions rather than changes requested. Once they are addressed, I think this is good to be merged!

src/js/main.js
mxmason: It's probably a micro-optimization, but: iterating with `forEach` is always more expensive than a `for...` loop, because `forEach` requires function calls; for loops do not.

SaptakS (approved):
Looks good to me. Merging.

PR #1248, opened 2021-04-23
Strip tags from raw string before sending to slugify

mxmason:
The `slugify` filter would pass entire html tags like `<code>` into its final output. This meant that the Checklist share links were much longer and complicated than they needed to be. For instance, `<code>title</code>` was converted to `-lesscodegreatertitleless/codegreater`

This change removes the `less`, `greater`, and `/` artifacts from the filter's output.


SaptakS (approved):
Regex logic looks good to me. Handles all the HTML stripping properly. Thanks for the PR! Merging!

PR #1260, opened 2021-04-29
Some light script cleanup

mxmason:
## Summary 
This PR:
- Catches an un-handled browser API support error
- Fixed a malformed string in a query selector

## Open questions
- [ ] Is the code related to email hrefs dead? `'[data-mailto]'` doesn't appear anywhere other than `main.js`. If it is dead, should we eliminate it from the repo altogether?

### DIFF
--- a/src/js/main.js
+++ b/src/js/main.js
@@ -75,7 +75,7 @@ function openLinkedCheckListItem() {
 	var hash = window.location.hash.substr(1);
 	var checklistItem =
 		hash.length > 0 &&
-		document.querySelector("[data-checklist-item-id=" + hash + "]");
+		document.querySelector('[data-checklist-item-id="' + hash + '"]');
 
 	if (checklistItem) {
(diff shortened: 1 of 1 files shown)


ericwbailey:
>Is the code related to email hrefs dead? '[data-mailto]' doesn't appear anywhere other than main.js. If it is dead, should we eliminate it from the repo altogether?

Yes, we should. We used to be able to use a `mailto`, so that data attribute handled some email-facing stuff. The email address stopped working, so we shifted to a contact form. 

Do you want to remove it in this PR, or should we do that in a subsequent one?

mxmason:
Let’s do it in a subsequent one. I’d like to make better decisions about when we serve scripts – e.g., only serve the code for the checklist on the checklist page; same for the contact page. As I do that, I’ll trim out the dead code.

ericwbailey (approved):
Heck yeah. Merged, then.

PR #1262, opened 2021-04-29
Load scripts only on pages that need them

mxmason:
## Summary
This PR:
- creates a `checklist.js` script, with all the code relevant to enhancing the checklist
- changes the `base.njk` template so that `checklist.js` and `contact.js` only load on relevant pages
- appends a version hash to `contact.js`
- deletes dead code previously used to enhance mailto links, as discussed in #1260
- deletes `main.js`, as there is currently no js that should run on every page


src/_includes/layouts/base.njk
```diff
@@ -29,10 +29,14 @@
...
 
-		<script async src="{{ '/' | url }}js/main.min.js?v={{getCurrentVersion.version}}"></script>
-    <script src="{{ '/' | url }}js/details-element-polyfill/details-element-polyfill.polyfill.min.js"></script>
-		<script src="{{ '/' | url }}js/table-of-contents/table-of-contents.min.js"></script>
-		<script src="{{ '/' | url }}js/contact.min.js"></script>
+    <script async src="{{ '/js/details-element-polyfill/details-element-polyfill.polyfill.min.js' | url }}"></script>
+		{% if page.url === '/checklist/' %}
+		<script src="{{ '/js/checklist.min.js?v=' + getCurrentVersion.version | url }}"></script>
+		{% endif %}
+		{% if page.url === '/contact/' %}
```
SaptakS, line 36: Any reason for not using `elif` instead of `if`. I don't think in any situation both the if conditions can match.

mxmason: Nothing special. As a developer, I just tend not to write control flows with `else if` – especially not when cases are mutually exclusive, like this

ericwbailey (approved):
I very much appreciate this.

Issue #1209, opened 2021-03-12
Add the level of the rule directly on the item listed


aguscha333:
Please review our Contributing guidelines:
https://a11yproject.com/contributing-guidelines/

## Summary
I think it would be useful to be able to quickly see the level of accessibility (A, AA, or AAA). Ideally, a webpage should be able to check them all off but it is also important to know what to prioritize to make the most impact sooner.

### Expected result

> Show the level (A, AA, or AAA) on each item in the accessibility checklist when the items are either collapsed or not.

### Actual result

> This is not available yet.

I would like to know if you think this is a good idea and how possible its implementation is as well. With some guidance, I'd be happy to take a crack at it.

ericwbailey:
Hey @aguscha333, I think this is a lovely idea! Some of the checklist items represent multiple WCAG success criterion, so this might get a little tricky. Do you think it'd be an average, or the highest applicable level?

As for implementation, [here's where the checklist content is stored](https://github.com/a11yproject/a11yproject.com/blob/main/src/_data/checklists.json). We could add a `wcag` entry for the JSON data, then update [the checklist partial](https://github.com/a11yproject/a11yproject.com/blob/main/src/_includes/checklist.njk) to output it.

aguscha333:
Hmm good question, I was not expecting them to relate to more than one level. I think if we are only showing one level I would show the lowest applicable level since it's the first level that people are going to try to cover so listing the highest one only could lead to the item being dismissed and left to do later. That being said, since there are only 3 levels I think that it could be possible to list them all like `A, AA` for example.

In case I want to contribute and add the change myself, I would need a bit of guidance in gathering the levels associated with each item in the checklist JSON. Each item has a `"wcag"` property that has the name of the guideline and then also a link. I haven't checked all of them but it would just be to open the link and copy the levels listed there?

I would add a property called `wcagLevels` and for simplicity the value could be just a string like `"A, AA"` or it could be an array like `["A", "AA"]` and then parsed some how. What do you think?

resource11:
I can take a crack at this, @ericwbailey

ericwbailey:
>I think if we are only showing one level I would show the lowest applicable level since it's the first level that people are going to try to cover so listing the highest one only could lead to the item being dismissed and left to do later. 

This makes sense to me!

>In case I want to contribute and add the change myself, I would need a bit of guidance in gathering the levels associated with each item in the checklist JSON.

[...]

PR #1268, opened 2021-05-06
Restructure checklist data

mxmason:
Unblocks some work in #940

## Summary
This PR restructures `checklist.json` data so that each checklist category includes its own `preface` commentary. This allows us to loop over the checklist data to render each category in the list, resulting in a dryer template file.

Also:
- `includes_checklist.njk` has been renamed to `_includes/checklist__task.njk`.

### Before

```json
{
  "content": [
    {
      "title": "Use plain language and avoid figures of speech, idioms, and complicated metaphors.",
      "checkboxId": "use-plain-language",
      "wcag": "3.1.5 Reading Level",
      "url": "https://www.w3.org/TR/UNDERSTANDING-WCAG20/meaning-supplements.html",
      "description": "Write content at <a href='https://datayze.com/readability-analyzer.php'>an 8th grade reading level</a>."
    },
  ]
}

```

### After

```json
{
  "content": {
    "preface": "Content is the most important part of your site.",
    "tasks": [
      {
        "title": "Use plain language and avoid figures of speech, idioms, and complicated metaphors.",
        "checkboxId": "use-plain-language",
        "wcag": "3.1.5 Reading Level",
        "url": "https://www.w3.org/TR/UNDERSTANDING-WCAG20/meaning-supplements.html",
        "description": "Write content at <a href='https://datayze.com/readability-analyzer.php'>an 8th grade reading level</a>."
      },
    ]
  }
}
```

PR #1270, opened 2021-05-09
feat: add WCAG levels

aguscha333:
Solves issue https://github.com/a11yproject/a11yproject.com/issues/1209

## Summary
This PR adds the WCAG level to each checklist item so it's easier to know what checks to prioritize when trying to make a site accessible.

## Notes / Questions
* Do you think that the placement of the new information is correct / the best way to display it?
I tried putting it in the part of the checklist item that is always visible but there there was a concern with responsiveness and the full `Level: AAA` was taking too much space.
* @ericwbailey you [mentioned](https://github.com/a11yproject/a11yproject.com/issues/1209#issuecomment-799887433) in the issue that some checklist items might represent multiple success criterions. At least from what I saw each checklist item has only one WCAG link and each of those links only had one WCAG level asociated.
That being said, I know for a fact that for example there are different contrast requirements for different levels but those don't seems to be covered in the checklist, just mention the one for level AA that is the minimum. Any thoughts on this? do you think that it's ok as I put it?
* There was one checklist item that linked to a WCAG technique instead of a "Understanding WCAG" rule and didn't have a level asociated to for that one I put `Level: doesn't apply`. I thought about putting nothing at all but then it would have looked like it was missing.

## Preview
image

### DIFF
--- a/src/_data/checklists.json
+++ b/src/_data/checklists.json
@@ -5,21 +5,24 @@
 			"checkboxId": "use-plain-language",
 			"wcag": "3.1.5 Reading Level",
 			"url": "https://www.w3.org/TR/UNDERSTANDING-WCAG20/meaning-supplements.html",
-			"description": "Write content at <a href='https://datayze.com/readability-analyzer.php'>an 8th grade reading level</a>."
+			"description": "Write content at <a href='https://datayze.com/readability-analyzer.php'>an 8th grade reading level</a>.",
+			"wcagLevel": "AAA"
 		},
 		{
 			"title": "Make sure that <code>button</code>, <code>a</code>, and <code>label</code> element content is unique and descriptive. ",
 			"checkboxId": "make-sure-interactive-content-is-unique",
 			"wcag": "1.3.1 Info and Relationships",
 			"url": "https://www.w3.org/TR/UNDERSTANDING-WCAG20/content-structure-separation-programmatic.html",
-			"description": "Terms like “click here” and “read more” do not provide any context. Some people navigate using a list of all buttons or links on a page or view. When using this mode, the terms indicate what will happen if navigated to or activated."
+			"description": "Terms like “click here” and “read more” do not provide any context. Some people navigate using a list of all buttons or links on a page or view. When using this mode, the terms indicate what will happen if navigated to or activated.",
+			"wcagLevel": "A"
 		},
 		{
 			"title": "Use left-aligned text for left-to-right (<abbr>LTR</abbr>) languages, and right-aligned text for right-to-left (<abbr>RTL</abbr>) languages.",
 			"checkboxId": "use-ltr-rtl",
 			"wcag": "1.4.8 Visual Presentation",
 			"url": "https://www.w3.org/TR/UNDERSTANDING-WCAG20/visual-audio-contrast-visual-presentation.html",
-			"description": "Centered-aligned or justified text is difficult to read."
+			"description": "Centered-aligned or justified text is difficult to read.",
+			"wcagLevel": "AAA"
 		}
 	],
 	"global": [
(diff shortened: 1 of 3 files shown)


mxmason:
Thank you so much for this pull request, @aguscha333! I love the idea of clarifying the WCAG level(s) associated with each checklist item. 

I did a major overhaul of the checklist data structure in #1268, and that blocks this PR for now. Once #1268 merges, I am happy to take responsibility for any merge conflicts that you experience here. 

Some thoughts this PR itself:
- I will leave @ericwbailey to take point on the placement and presentation of this new information
- I think we could expand the data to associate more WCAG criteria with each checklist item, where appropriate. This could be a separate PR; the current work is a great start
- I think `Level: doesn't apply` is fine for the cases where there's not a relevant success criterion

src/js/checklist.js
```diff
@@ -8,6 +8,25 @@
 	Element.prototype.matches = Element.prototype.msMatchesSelector;
 }
 
+var checklistCategoryExpanded = {
```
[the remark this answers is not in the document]
mxmason: Yes, that was my mistake! I just fixed up the history to take out the irrelevant changes! Sorry about that?!

src/_data/checklists.json
```diff
@@ -454,101 +508,64 @@
...
 				"url": "https://www.w3.org/WAI/WCAG21/Understanding/animation-from-interactions.html",
-				"description": "Remove animations when the “reduce motion” setting is activated. If an animation is necessary to communicate meaning for a concept, slow its duration down."
+				"description": "Remove animations when the “reduce motion” setting is activated. If an animation is necessary to communicate meaning for a concept, slow its duration down.",
+				"wcagLevel": "AAA"
 			}
 		]
 	},
 	"color contrast": {
-		"preface": "<a class=\"c-preface__link\" href=\"{{ '/posts/2015-01-05-what-is-color-contrast/' | url }}\">Color contrast</a> is how legible colors are when placed next to, and on top of each other.",
-		"tasks": [
```
aguscha333, line 477: is the removal of this tasks correct?

mxmason: Another mistake from when I was working with the diff. Fixed :)

src/_data/checklists.json
```diff
@@ -480,75 +537,85 @@
...
 				"checkboxId": "selection-contrast",
 				"wcag": "1.4.3 Contrast",
 				"url": "https://www.w3.org/TR/UNDERSTANDING-WCAG20/visual-audio-contrast-contrast.html",
-				"description": " Is the color contrast you set in your <a href='https://developer.mozilla.org/en-US/docs/Web/CSS/::selection'><code>::selection</code> CSS declaration</a> sufficient? Otherwise someone may not be able read it if they highlight it. "
+				"description": " Is the color contrast you set in your <a href='https://developer.mozilla.org/en-US/docs/Web/CSS/::selection'><code>::selection</code> CSS declaration</a> sufficient? Otherwise someone may not be able read it if they highlight it. ",
+				"wcagLevel": "AA"
 			}
 		]
 	},
-	"mobile and touch": {
```
SaptakS, line 522: @mxmason I think this is another thing that got carried over during resolving merge conflicts.

mxmason:
Thanks for these notes, @scottaohara!

@aguscha333, I'd be happy to implement the content changes Scott has pointed out, if you'd like. I think these changes could either go into this PR, or into a separate one dedicated to the rewrite.

aguscha333:
@mxmason thanks for offering. If you don't mind I'd like to try to implement the changes myself to keep contributing to the repo. Those changes do touch more technical subjects about WCAG itself so what I will need is help with a thorrow review since I only recently started learning about accessibility so I could have some conceptual errors in the changes.
I do think that it's better to do it in a separate PR since the new changes have nothing to do with the initial purpose of this PR. I can create a branch off of this one to get minimal conflicts.

mxmason:
@aguscha333 sounds good to me! feel free to tag me directly if you need anything!

SaptakS:
Sounds like a good plan to me too. I guess we are tracking it in https://github.com/a11yproject/a11yproject.com/issues/1277 now (thanks @mxmason for creating the issue!). In that case, according to me, this PR is ready to be merged. cc @scottaohara @ericwbailey

aguscha333:
@ericwbailey @mxmason Tried implementing #1277 but going through the suggestions there are a lot that are more open ended and would require some more decision making from someone with more knowledge on the topic. I'm glad to keep contributing in the future but I think I need to sit this one out since I am still a newbie with a11y.

scottaohara:
if you think more work needs to go into correctly citing the WCAG SCs, then I would suggest you simply comment out the SCs for the checklist items where there is not a simple swap out of the errant SC for the correct one.

Then additional thought can go into the items that need more nuance, without continuing to cite inaccurate SCs while that work is being done.

aguscha333:
@scottaohara @ericwbailey sorry it took me soo long to get back to this.

Unfortunately comments are not allowed in JSON files so I had to remove them instead.

I took the removal approach to most of the items mentioned by @scottaohara since most of the comments looked like were meant to open up discussions rather than this should be exactly like this or that.

If anyone wants to jump in and do something different for the ones I removed, totally cool with that.

PR #1279, opened 2021-05-19
Adds a11ysd

schne324:


ericwbailey (changes requested):
Thanks for your submission, @schne324! Would you mind moving this to be in alphabetical order to match the rest of the entries? Once that's set I'll happily merge it.

schne324:
@ericwbailey it didn't appear that everything was quite in alphabetical order but I've placed a11ySD right before a11ySEA. Let me know if I need to move it somewhere else

ericwbailey (approved):

ericwbailey:
Thanks for doing that, @schne324. I'll take a look at the order as well.

Issue #1282, opened 2021-05-25
Checklist item clarifications: Questions/comments on 2 checklist items


MicheleAWilliams-A11y:
Hi there, 2 questions/comments in the checklist:

### Appearance > Double-check that good proximity between content is maintained

  The "straw test" mentioned in the description is very useful; however, I didn't see anything else in the checklist representing 1.3.3's sensory guidance except the related "color only" guidance of 1.4.1 - Use of Color.  
  
  I know the checklist isn’t all-inclusive of WCAG 2.1 AA but I think it's worth noting other non-color characteristics such as _spatial directions_ (e.g., “column on the left" which gets impact by browser zoomed pages and of course screen readers/braille displays) and _audio cues_ (e.g., “record after the tone" which needs visual confirmation for deaf and hard of hearing folks). 

Just seemed the additional sensory information was a good low-hanging fruit guideline to include along with color since it is a formal WCAG item as well as not always remembered (or known).

### Global Code > Remove session timeouts   

The description text rightly clarifies this is about adjustable timing and starts with “If you cannot”. However, given people might (likely often) take the top point as the guideline without further reading (as I actually just experienced in working with Webflow folks on their checklist), I wondered if there was a reason to be so direct in "removing timeouts" rather than, say, "warning and extending timeouts"? 

Not to say timeouts are the greatest but having worked with a bank, for instance, I can attest these timeout discussions get touchy (heavy sigh) and saying "remove them" would be considered blasphemy. Just wanted to point out this checklist item may be perceived as the place that definitively says "accessibility requires removing timeouts" in case that wasn't your intent. 

Just some thoughts as an observer who saw the checklist in use by newbies. Happy to have more discussion. Thanks!

ericwbailey:
@MicheleAWilliams-A11y I think both of these are excellent points. Is this something you'd like to help author the updates for?

MicheleAWilliams-A11y:
@ericwbailey - Awesome, yes indeed! I'll start some notes on my end and let me know where I should put them for consideration.

FYI, Webflow held their timeout guideline for later since that's not a current feature but here's what they put for sensory after some of my feedback: https://webflow.com/accessibility/checklist/task/add-descriptions-to-sensory-based-instructions

ericwbailey:
Closed via https://github.com/a11yproject/a11yproject.com/pull/1290. Congratulations on your first contribution to the project, @MicheleAWilliams-A11y!

PR #1290, opened 2021-06-08
Changed wording for session timeouts

MicheleAWilliams-A11y:
Hi all, these are my updates per Issue #1282. Please have a look and let me know if anything else is needed!


ericwbailey (approved):
Wonderful! Nice use of fancy quotes, too 🙂

PR #1296, opened 2021-06-18
Use dart-sass to compile scss

mxmason:
`gulp-sass` uses `node-sass` to compile by default; this override changes that.

Changing the compiler also required me to fix some malformed scss.


SaptakS (approved):
Looks good to me.

PR #1298, opened 2021-06-25
Chore: align to Node 16

mxmason:
Apple's new M1 devices only work with Node 15+. Given that Node 16 is _almost_ at the LTS release stage, I think it's safe to align our requirements with Node 16 to save Mac-using contributors from experiencing any pain.

This PR also includes a smol migration to `gulp-sass@5`, which was blocking us from migrating to Node 16. Bonus: the new release of `gulp-sass` no longer includes `node-sass`, so now our installation is ~14mb lighter! 🎉


SaptakS:
The CIs seem to to be failing. Probably because of a mismatch in the OS version and dependency versions?

mxmason:
@SaptakS looks like I was able to fix this by recreating the lock file so it has the right version of `fsevents`

PR #1301, opened 2021-06-29
Use directory data files to deduplicate post front matter

mxmason:
## Summary

This PR utilizes Eleventy's template data files, and Eleventy computed data, to DRY up the front matter required for both posts and announcements. Before, each post required explicit `layout` and `eleventyNavigation` keys; now they do not. 

Hopefully, this will reduce some repeated effort for all of us, and also eliminate some overhead for our contributors, who should not need to know about this front matter.


src/posts/posts.11tyData.js
```diff
@@ -0,0 +1,18 @@
...
+		},
+		tags: ({ category, tags }) => {
+			if (tags) {
+				return tags;
+			} else {
+				return [category.replace(ALPHABETICAL_RE, "").toLowerCase()];
+			}
+		},
+	},
+};
```
mxmason, line 18: 1️⃣ 🔍 

This JavaScript file sets the layout of every `post` template to `'post'`. It also reads the front matter-supplied `category` and `title` to set `eleventyNavigation` data.

src/announcements/announcements.11tyData.js
```diff
@@ -0,0 +1,12 @@
...
+	category: "Announcement",
+	layout: "post",
+	tags: ["announcements"],
+	eleventyComputed: {
+		eleventyNavigation: {
+			key: ({ category }) => category,
+			title: ({ title }) => title,
+		},
+	},
+};
```
mxmason, line 12: 2️⃣ 🔍 

Same idea as 1️⃣ , except it hard-codes more values. Now, each announcement should only need a `title` and a `description` in its front matter.

mxmason (commented):
@ericwbailey @SaptakS, I left some specific comments about these new directory data files 🙇🏻

src/posts/posts.11tyData.js
```diff
@@ -0,0 +1,18 @@
+const ALPHABETICAL_RE = new RegExp("[^a-zA-Z0-9 -]");
```
SaptakS, line 1: This is basically to check if the category has anything apart from alphabets, digits, space and "-", right? The constant name kind of threw me off a bit. Are there specific categories because of which this regex replace was needed?

mxmason: Hey, @SaptakS!

1. yes, that’s what this does
2. For one example, we have a `How-to` category. The corresponding tag as of now is `howto`

SaptakS:
@mxmason A particular post is looking unstyled: https://deploy-preview-1301--thea11yproject.netlify.app/posts/2020-09-19-how-to-write-accessible-forms/ I checked both locally and in deploy-preview. Not sure exactly why but seems the post layout is not being applied.

mxmason:
Thanks for catching that, @SaptakS. My file system is not case-sensitive, so it wasn't catching the problem (that the js data file must be called `11tydata`, not `11tyData`. Deploy preview looks fine now! 

I also renamed that regex and added a comment :)

SaptakS (approved):
Looks good to me.

PR #1309, opened 2021-07-07
Add recommended vscode settings & extensions

mxmason:
The files in `.vscode` will:
- Recommend to new users some extensions related to our dev tools (currently: editoconfig, joshing, nunjucks, stylelint)
- Disable validating of CSS and SCSS files, because stylelint does that for us


ericwbailey (approved):
Oh, this is nice. I didn't know VS Code could do this.

PR #1331, opened 2021-08-14
Fix heading text overflow on small screen on the sponsorship page

### LINKED ISSUE #1330: Text overflow on the sponsorship page when using  device with smaller screen
## Summary

1. Navigate to the [sponsorship page](https://www.a11yproject.com/sponsorship/) using chrome
2. Open the browser dev tools 
3. Toggle the device tool bar to mobile
4. Select `Iphone 5/SE` as the device

### Expected result

The text in the `h1` element (Sponsorship) should not overflow.

### Actual result

The text in the `h1` element (Sponsorship) overflows as shown in the image below.
image

### Additional Information

1. What device were you using?  Laptop
2. Operating System version. Ubuntu 20.04
3. Browser(s) and version(s). Chrome Version 92.0.4515.131 (Official Build) (64-bit)

nibble0101:
## PR description

This PR will fix  text overflow on small screens on the sponsorship page.

## Before

Currently, this is what the `h1` element on the sponsorship page looks like  when viewed on small screens.
image

## After

After merging this PR,  the `h1` element on the sponsorship page will look like in the image below when viewed on small screens 
image

## PR

Fixes #1330

### DIFF
--- a/src/css/components/content/_headings.scss
+++ b/src/css/components/content/_headings.scss
@@ -4,6 +4,7 @@
 	h2,
 	h3 {
 		line-height: $line-height-tight;
+		overflow-wrap: break-word;
 
 		// Breakpoints
 		@include mappy-bp(lap-medium) {


nibble0101:
@ericwbailey  I am not really very  familiar with sass. I am still trying to make sense of the styling.  Let me know what you think of this fix.

SaptakS (commented):
This in general looks good to me. Right now, seems like the announcements page is the only one missing `c-content` unlike all other pages. Can you also probably add there [here](https://github.com/a11yproject/a11yproject.com/blob/main/src/announcements.njk#L14) as part of the PR, so that it fixes all the pages?

It seems like only `.c-content--generic h1` are right now affected, but I don't think there is any big issue applying this to all according to my review.

ericwbailey (approved):

ericwbailey:
@nibble0101 This is great, and I think your Sass change is placed in exactly the right place. If you're not comfortable updating the PR with @SaptakS' fix we can do a followup PR.

nibble0101:
> @nibble0101 This is great, and I think your Sass change is placed in exactly the right place. If you're not comfortable updating the PR with @SaptakS' fix we can do a followup PR.
 
@ericwbailey  It is okay, I can work on @SaptakS's suggestion before merging. I intend to look into it  by the end of the week.

Issue #1344, opened 2021-08-31
Article: Making links accessible


Kodecheff:
Hi @ericwbailey 🙂, I have a post idea I want to write about - `making links accessible`. Is this a good idea to write on? 

## Description

> Link is one of the primary concept in web development. The article provides quick tips on making links accessible to everyone. 

## Would you like to write the article or request that it be written?

> I would like to write the article

ericwbailey:
@Kodecheff Could you please provide an outline for us to review?

Kodecheff:
Sure @ericwbailey 
Links are one of the primary elements in web development. Links can be made accessible by making them clear, descriptive and purposeful out of context. This helps break accessibility barriers for both screen readers users and clear sighted users. 
* Link texts such as `more`, `click here` are ambiguous and not appropriate. Rather a concise link text so users can know the exact destination of the link should be provided.
* The `aria-label` attribute should be used, because it provides a descriptive link text for screen readers users. 
* Links should not be made to look like a button, to avoid confusion.
* Links text are best not repeated to avoid disorientation for users with cognitive disability. 

There are several mistakes web authors make that creates a barrier for accessibility on their websites or applications in regards to hypertext links. This article will provide a guide to help create an accessible website or application.

Thank you.

ericwbailey:
@Kodecheff This sounds good. Please make sure you read and adhere to our [Content Style Guide](https://www.a11yproject.com/content-style-guide/) before submitting the post for review.

Daniel-Dragon:
I'd like to grab this and run with it if that's alright. I have a basic outline started, but might add more if I think of anything important.

- **How Accessibility Software (Or say Screen Reader?) Uses Links** -- Discuss the WHY this is important to consider, show how they list all links ETC. I think getting the WHY helped to explain why the suggestions are the way they are.
- **Avoid Redundancies** -- Avoid `link to` or `click here` because it's implied by the elements role.
- **Avoid Link Blocks** -- It's exhausting too have too much text linked. Keep links to < 100 characters long if possible. Can even go into some strategies on how to expand clickable area to a container using CSS (a strategy I've used with some success)
- **Avoid Unexpected New Windows** -- Need to use text or an icon with some alt text to alert the user that this link will open a new window to avoid confusing.
- **Links on Images** -- the image should have alt that describes the link location or the action taken, rather than describing the content of the image in this case.
[...]

PR #1346, opened 2021-09-02
Article: making links accessible

Kodecheff:
@ericwbailey please check review

This addresses #1344

### DIFF
--- a/src/posts/2021-09-08-making-links-accessible.md
+++ b/src/posts/2021-09-08-making-links-accessible.md
@@ -1,0 +1,70 @@
+---
+title: Making links accessible
+description: Tips on how to make links accessible.
+category: How to
+author: Pascal Akunne
+date: 2021-09-08
+further_reading:
+  - title: Links and Hypertext
+    url: https://webaim.org/techniques/hypertext/
+    source:  WAIM Web Accessibility In Mind
+    year: 2021
+  - title: Link Purpose (Link only)
+    url: https://www.w3.org/TR/UNDERSTANDING-WCAG20/navigation-mechanisms-link.html
+    source:  WCAG 2.0
+    year: 2016
+  - title: Links
+    url: https://accessibility.umn.edu/what-you-can-do/start-7-core-skills/links
+    source: University of Minnesota
+    year: 2021
+tags:
+  - howto
+---
+
+Links are one of the primary elements in web development. Links can be made accessible by making them clear, descriptive and purposeful out of context. This helps break accessibility barriers for both people with visual disability and clear sighted, as well as people with cognitive disability. Links should be made obvious so people can understand what is means. By default, links are underlined by the browser. 
+
+
+## Guidelines for making links accessible
+
+### Don'ts
+
+- Careless use of ambiguous link texts like `Click here`, `More...`, `Link` are not appropriate and should be avoided. Rather, a text that describes the purpose of that link should be used. 
+- Do not use a URL as your link text. This makes it difficult for most screen readers to understand, they read each letter of the URL. 
+- Link texts are best not repeated to avoid disorientation for users with cognitive disability. 
+- Links should not be made to look like a button, to avoid confusion. Links navigate to webpages while buttons perform other actions like submitting forms. 
+- The `href` attribute should not be empty. Some developers use the pound sign (`#`) as the value for the `href` attribute. There's no point in providing a link without a destination. 
+
+### Dos
+
+- The `aria-label` attribute should be used. Because it provides a descriptive link text for people that are visually impaired.
+```
+<a href="https://w3schools.com/html" aria-label="Read more on HTML" >Read more</a>
+```
+- For easy readability, link texts should be underlined. Although, traditionally, link texts are underlined with a color by the browser.
+- Use the `cursor: pointer` declaration on your CSS. This helps to identify a link on a webpage using the `a:hover` style. 
+- Links should be made easy to navigate to and select using the `tab` key and `enter` key respectively. 
+- Link texts should be short and concise. Using a paragraph or an entire sentence is unnecessary and frustrating. 
+- Use `skip` link to skip over a long list of links. Some websites use `skip navigation` to skip to the main content. 
+```
+<header>
+  <a href="#main-content" aria-label="Skip to main content" >Skip navigation</a>
+
+  <nav>
+    <ul>
+      <li><a href="home.html">Home</a></li>
+      <li><a href="contact.html">Contact</a></li>
+      <li><a href="about.html">About</a></li>
+    </ul>
+  </nav>
+</header>
+
+<div id="main-content">
+  <!-- Content goes in here -->
+</div>
+```
+
+## Summary
+
+You can improve both the accessibility and usability of your website or application by making the links concise, descriptive, and purposeful out of context.
+
+There are several mistakes web authors make that create a barrier for accessibility on their websites or applications in regards to hypertext links. Following this guidelines will help you to create an accessible website or application.
--- a/src/_data/authors.json
+++ b/src/_data/authors.json
@@ -151,6 +151,12 @@
 			"biography": "Founder of MadHackerHaus and Rolled Up Sleeves, LLC. Advocate for accessibility in developing the internet and mobile technology."
 		},
 		{
+			"name": "Pascal Akunne",
+			"website": "https://kodecheff.netlify.app",
+			"photo": "https://pbs.twimg.com/profile_images/975451645462106112/_h6vsUym_400x400.jpg",
+			"biography": "JavaScript (Node.js & React.js) Developer and Technical Writer"
+		},
+		{
 			"name": "Rachel R. Vasquez",
 			"website": "http://rachievee.com",
 			"photo": "https://avatars0.githubusercontent.com/u/10246844?s=460&v=4",


src/posts/2021-09-08-making-links-accessible.md
```diff
@@ -0,0 +1,70 @@
...
+
+### Dos
+
+- The `aria-label` attribute should be used. Because it provides a descriptive link text for people that are visually impaired.
+```
+<a href="https://w3schools.com/html" aria-label="Read more on HTML" >Read more</a>
+```
+- For easy readability, link texts should be underlined. Although, traditionally, link texts are underlined with a color by the browser.
+- Use the `cursor: pointer` declaration on your CSS. This helps to identify a link on a webpage using the `a:hover` style. 
+- Links should be made easy to navigate to and select using the `tab` key and `enter` key respectively.
```
scottaohara, line 45: again this is default behavior, so maybe revise this wording to indicate that... as there shouldn't be anything special a developer would need to do here.

scottaohara (changes requested):
this could use some additional work to provide more robust examples and suggestions for making accessible links.  The summary could stand to repeat some of the best practices, rather than repeating the problem statement and then indicating the reader read the article to find out more.

i've provided some specific commentary to start you off.

Kodecheff:
@scottaohara thank you for your suggestions. I will look into it right away.

Kodecheff:
Hi @scottaohara, please review my update.

src/posts/2021-09-08-making-links-accessible.md
```diff
@@ -0,0 +1,74 @@
...
+---
+
+Links are one of the primary elements in web development. Links can be made accessible by making them clear, descriptive and purposeful out of context. This helps break accessibility barriers for both people with visual disability and clear sighted, as well as people with cognitive disability. Links should be made obvious so people can understand what is means. By default, links are underlined by the browser. 
+
+
+## Guidelines for making links accessible
+
+### Don'ts
+
+- Careless use of ambiguous link texts like `Click here`, `More...`, `Link` are not appropriate and should be avoided. Rather, a text that describes the purpose of that link should be used.
```
scottaohara, line 31: don't need the word "careless" as that makes an unnecessary negative assumption of the person writing the text.

additionally "are not appropriate and should be avoided" also lacks context.  For example "more options", "learn more", when used in the context of a paragraph are acceptable to meet 2.4.4.  If you are going to say they're inappropriate, you need to provide context as to why.

src/posts/2021-09-08-making-links-accessible.md
```diff
@@ -0,0 +1,74 @@
...
+
+## Guidelines for making links accessible
+
+### Don'ts
+
+- Careless use of ambiguous link texts like `Click here`, `More...`, `Link` are not appropriate and should be avoided. Rather, a text that describes the purpose of that link should be used. 
+- Do not use overly long URLs as your link text. This makes it difficult for most screen readers to understand. Short and concise URLs such as twitter.com, github.com/kodecheff may be accepted. This would help users identify the text strings as links if the situation called for it.
+- Link texts are best not repeated to avoid disorientation for users with cognitive disability. 
+
+- The `href` attribute should not be empty. Some developers use the pound sign (`#`) as the value for the `href` attribute. There's no point in providing a link without a destination.
```
scottaohara, line 35: again, some context here.... as while not at all ideal, there are cases where developers who have successfully implemented Single Page Application routing may have links that render in the DOM as `<a href="#">`, but when activated they do actually navigate users to the intended web page.  So generally speaking, yes this is correct.  But per the above, there is nuance that may need to be explained as to why, even in the situation I indicated, a link with a `href=#` value could pose some issues (though not necessarily solely accessibility related).

src/posts/2021-09-08-making-links-accessible.md
```diff
@@ -0,0 +1,74 @@
...
+
+- Careless use of ambiguous link texts like `Click here`, `More...`, `Link` are not appropriate and should be avoided. Rather, a text that describes the purpose of that link should be used. 
+- Do not use overly long URLs as your link text. This makes it difficult for most screen readers to understand. Short and concise URLs such as twitter.com, github.com/kodecheff may be accepted. This would help users identify the text strings as links if the situation called for it.
+- Link texts are best not repeated to avoid disorientation for users with cognitive disability. 
+
+- The `href` attribute should not be empty. Some developers use the pound sign (`#`) as the value for the `href` attribute. There's no point in providing a link without a destination. 
+
+### Dos
+
+- The `aria-label` attribute should be used where appropriate. Because it provides a descriptive information for people that are visually impaired. If used within a paragraph, it could cause a poor experience as the text of the paragraph could sound unnecessarily awkward when reading all at once.
```
scottaohara, line 39: thanks for updating this with some of the information i commented on, but we should still be more clear here.

* 'used where appropriate' is still a bit lacking.  With the updated text we allude to an example of where this isn't appropriate to do, and then there's a HTML example of a link without context where this could be appropriate.  If we want this to be a useful article of dos / donts, then we should provide as many details and examples as possible, otherwise the remaining ambiguity is what people will take away from the article.
* If not this bullet directly, a second bullet needs to be made to indicate that is using an `aria-label` on a link, then it MUST also meet 2.5.3 Label in Name standards.  Again, a good example vs bad example would be beneficial here.

src/posts/2021-09-08-making-links-accessible.md
```diff
@@ -0,0 +1,74 @@
...
+```
+<a href="https://w3schools.com/html" aria-label="Read more on HTML" >Read more</a>
+```
+- For easy readability, link texts should be underlined. Although, traditionally, link texts are underlined with a color by the browser.
```
scottaohara, line 43: Where might there be exceptions to this rule?  As there are some, and while this rule should generally be followed, indicating there are exceptions would again make this more relatable to readers who might otherwise interpret this as too draconian of a stance to take.

src/posts/2021-09-08-making-links-accessible.md
```diff
@@ -0,0 +1,74 @@
...
+- By default, the browser uses the `cursor: pointer` declaration on your CSS. You can use the `a:hover` or `a:focus` style to apply other CSS styles such as color, background-color, or text-decoration, to help identify links on the browser.
```
scottaohara, line 44: Again, do we even need to tell people what the browser does by default here?  More often than not, developers are adding `cursor: pointer` to interactive elements which aren't links, rather than removing it from links.  

Might also be good to note that when modifying the color/background, focus styles, etc, that there are color contrast and focus visible rules that would need to be adhered to.

src/posts/2021-09-08-making-links-accessible.md
```diff
@@ -0,0 +1,74 @@
...
+
+### Dos
+
+- The `aria-label` attribute should be used where appropriate. Because it provides a descriptive information for people that are visually impaired. If used within a paragraph, it could cause a poor experience as the text of the paragraph could sound unnecessarily awkward when reading all at once.
+```
+<a href="https://w3schools.com/html" aria-label="Read more on HTML" >Read more</a>
+```
+- For easy readability, link texts should be underlined. Although, traditionally, link texts are underlined with a color by the browser.
+- By default, the browser uses the `cursor: pointer` declaration on your CSS. You can use the `a:hover` or `a:focus` style to apply other CSS styles such as color, background-color, or text-decoration, to help identify links on the browser. 
+- If the only way to access a link is with a mouse, the link is unusable by keyboard users. Traditionally, links can be easily to navigated to and selected using the `tab` key and `enter` key respectively. This helps keyboard users to easily access your website. Except disallowed using [JavaScript event handlers](https://webaim.org/techniques/javascript/eventhandlers).
```
scottaohara, line 45: Is this really a "do?"   As this is the default behavior, wouldn't this be better under a don't with the text indicating "Do not remove the ability for links to be navigated to via keyboard...." etc?

src/posts/2021-09-08-making-links-accessible.md
```diff
@@ -0,0 +1,74 @@
...
+<a href="https://w3schools.com/html" aria-label="Read more on HTML" >Read more</a>
+```
+- For easy readability, link texts should be underlined. Although, traditionally, link texts are underlined with a color by the browser.
+- By default, the browser uses the `cursor: pointer` declaration on your CSS. You can use the `a:hover` or `a:focus` style to apply other CSS styles such as color, background-color, or text-decoration, to help identify links on the browser. 
+- If the only way to access a link is with a mouse, the link is unusable by keyboard users. Traditionally, links can be easily to navigated to and selected using the `tab` key and `enter` key respectively. This helps keyboard users to easily access your website. Except disallowed using [JavaScript event handlers](https://webaim.org/techniques/javascript/eventhandlers). 
+- Link texts should be short and concise. Using a paragraph or an entire sentence is unnecessary and frustrating. 
+- Use `skip` link to skip over a long list of links. Some websites use `skip navigation` to skip to the main content. 
+```
+<header>
+  <a href="#content" aria-label="Skip navigation">Skip navigation</a>
```
scottaohara, line 50: why is there an `aria-label` here?

src/posts/2021-09-08-making-links-accessible.md
```diff
@@ -0,0 +1,74 @@
...
+  <nav>
+    <ul>
+      <li><a href="home.html">Home</a></li>
+      <li><a href="contact.html">Contact</a></li>
+      <li><a href="about.html">About</a></li>
+    </ul>
+  </nav>
+</header>
+
+<div id="content">
```
scottaohara, line 61: would expect this to be a `<main>` with a `tabindex=-1`.   Or better yet, provide a `<h1 id=whatever tabindex=-1>...</h1>` within the main and have the skip link jump to that.

scottaohara (changes requested):

src/posts/2021-09-08-making-links-accessible.md
```diff
@@ -0,0 +1,74 @@
...
+    year: 2016
+  - title: Links
+    url: https://accessibility.umn.edu/what-you-can-do/start-7-core-skills/links
+    source: University of Minnesota
+    year: 2021
+tags:
+  - howto
+---
+
+Links are one of the primary elements in web development. Links can be made accessible by making them clear, descriptive and purposeful out of context. This helps break accessibility barriers for both people with visual disability and clear sighted, as well as people with cognitive disability. Links should be made obvious so people can understand what is means. By default, links are underlined by the browser.
```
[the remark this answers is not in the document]
Kodecheff: Hi @scottaohara, thanks for the feedback. "them" in the above context means their"texts" and how they are visually presented on the browser. So yeah, I'll say both.

Kodecheff:
Hi @scottaohara please review my update.

Kodecheff:
Hi @scottaohara, is the article good to go? 🙂

scottaohara:
@Kodecheff I will review when i have time.  please be patient as this is a volunteer effort on my part.

Kodecheff:
So sorry, I was over excited 😄. Please take your time.

Issue #1349, opened 2021-09-29
Safari 15.0 inherits harsh/failing UI Theme-color


coburnicus:
## Summary

The updated Safari (15.0) allows for the app UI color to be set by the website. Currently on a11yproject.com this is a pretty harsh Magenta with white type. That would seem to be pretty harsh to look at, and seems to fail WCAG contrast tests.

### Expected result

Something the passes WCAG contrast tests while adhering to A11yProject branding.

### Actual result

A lot of hot hot pink.
![pretty_in](https://user-images.githubusercontent.com/743929/135323126-777a8467-2a1c-4380-85d2-5932c491ea9f.png)

### Additional Information

This would seem to affect safari on desktop and mobile, and chrome on android. 
I'm created a fork with a potential fix. Hope I am being helpful. I really appreciate the work people put into this website. I kind of suck at git, so bear with me.

ericwbailey:
Closed via https://github.com/a11yproject/a11yproject.com/pull/1350

PR #1350, opened 2021-09-29
Update meta theme-color code

coburnicus:
Added some meta theme-color code that will effect the new safari-ui. This make the light-ui the light pink (LA Vibes), and the dark-ui a dark grey (nightmare!?). I thought the super bright pink with white font (current safari UI) made it hard to use, and probably failed contrast tests.

This is in response issue  that I created. #1349

### DIFF
--- a/src/_includes/layouts/base.njk
+++ b/src/_includes/layouts/base.njk
@@ -5,7 +5,7 @@
 		<meta http-equiv="x-ua-compatible" content="ie=edge" />
 		<meta name="viewport" content="width=device-width, initial-scale=1.0" />
 		<meta name="theme-color"
-      content="#ebf6fd"
+      content="#f6f6f6"
       media="(prefers-color-scheme: light)">
 		<meta name="theme-color"
       content="#333536"


ericwbailey (changes requested):
Hey! Thank you for catching this. I agree, the experience it creates is poor. I'm personally not a huge fan of Safari doing this with theme color for contrast reasons, as well as digital literacy and breaking the consistency of the experience across different websites.

Because of that, could I trouble you to remove the `meta` declarations altogether?

coburnicus:
Hi!
I think if I remove what I added, it defaults back to the magenta (for some reason safari is doing this on its own). So maybe the solution is to make the meta declarations very default-looking? Like a light-grey, so that it looks like the old defaults?

coburnicus:
Via Jen Simmons session on Safari 15:

> For sites that don’t provide a color, Safari will figure out a color to use. Sometimes Safari matches the background color of the web page. Other times, it matches the header background. To provide your choice for the toolbar color, put it in the HTML head using a meta tag with name=“theme-color,” and use the content attribute to specify the color. 

For some reason Safari has decided day-glow pink is what a11yproject browser should be ¯\_(ツ)_/¯.  I would say setting the colors would create the most usable experience for folks. I updated my commit to try and replicate the previous safari's light and dark modes.

**Light**
white-smoke

**Dark**
carbon

Issue #1377, opened 2021-11-11
Bug: Color contrast of <code> text in recommended reading link titles when hovered


larryhudson:
Thank you for reporting a bug. Please fill out the following information:

## Bug description

> Please describe the bug you encountered. This includes the bug itself and where you encountered it.

When the recommended reading links at the bottom of a post are hovered, the title text becomes white on a red background. If the title includes a `<code>` snippet, the `<code>` text also becomes white. Because `<code>` snippets have a grey background, the white on gray combination is not accessible. You can see an example of this on this page, in the 'next post' link at the bottom of the post:

https://www.a11yproject.com/posts/2015-05-29-getting-started-with-nvda/

Here's a screenshot of the 'next post' link when hovered:
![Screenshot of next post link. The code snippet has white text on a grey background, which is not accessible](https://user-images.githubusercontent.com/10627494/141375589-a33984ca-369d-4ea6-b440-acef17cf780d.png)

## Expected outcome

> Please describe what you would expect to happen if the bug did not exist.

Add SCSS that ensures the `<code>` text stays gray, rather than setting the color to white.

## Steps to reproduce

> Please describe the steps we can take to reproduce the bug. This will help us locate and fix it.

See 'next post' link at the bottom of this post:
https://www.a11yproject.com/posts/2015-05-29-getting-started-with-nvda/

###  What device were you using when you encountered the bug?

Desktop PC.

### What operating system were you using? What version is the operating system?

Windows 10.

### What web browser were you using? What version is the web browser?

Firefox Developer Edition 95

## Additional information

> Is there anything else we should know about?

I'm making a pull request that solves this issue. Will link through to this issue when I create it.

## Terms

- [x] I have read, and agree to follow this project's [Code of Conduct](https://www.a11yproject.com/code-of-conduct/)

PR #1378, opened 2021-11-11
Fix colour contrast issue on recommended reading links

larryhudson:
This pull request fixes [issue 1377](https://github.com/a11yproject/a11yproject.com/issues/1377).

### DIFF
--- a/src/css/components/content/_links.scss
+++ b/src/css/components/content/_links.scss
@@ -27,6 +27,7 @@ a {
 	// Style guide: Components.post.links
   a {
 		// States
+		&:hover,
 		&:focus {
 			// Text-level formatting
 			code {


SaptakS:
Hi. Thanks for the fix!! This looks good to me. However, seems the focus state already handles the error pretty nicely. 
image

This is the code for handling the focus state: https://github.com/a11yproject/a11yproject.com/blob/ae67e9d922dc155b61e89325444a8a426e4003d2/src/css/components/content/_links.scss#L30-L36 Maybe we can just add `&:hover` also in that anchor tag states? Will give it a more consistent look according to me.

larryhudson:
Awesome, that's a great idea. That'll make the contrast even better, and it'll fix all links with `<code>` snippets, not just the recommended reading links.

SaptakS (approved):
Looks great! Thanks!

Issue #1382, opened 2021-12-01
Bug: Incorrect RSS URL on the 'follow' page 


a11yjc:
**Follow page:** https://www.a11yproject.com/follow/
**Current RSS URL for hyperlink 'a RSS feed':** https://www.a11yproject.com/follow/ - This links to it's own page, which is incorrect. Code is `<a href="">a RSS feed</a>`
**'a RSS feed' hyperlink should be:** https://www.a11yproject.com/feed/feed.xml - This links correctly to the RSS output. Code should be `<a href="https://www.a11yproject.com/feed/feed.xml">a RSS feed</a>`

PR #1383, opened 2021-12-01
Added correct URL for the 'a RSS feed' hyperlink on the follow page

a11yjc:
To resolve: https://github.com/a11yproject/a11yproject.com/issues/1382


a11yjc (commented):
Adding correct Eleventy attribute instead of hardcode.

ericwbailey (approved):
Lovely! Thank you for this 🎉

Issue #1368, opened 2021-10-13
Dates in the post filenames don't always match the day they're published


ericwbailey:
The date metadata in a post filename (i.e. `2021-06-07-`) doesn't always match the date the post is published. Oftentimes this is because of the editorial process, which requires edits that can span one or more days.

I'm thinking we remove them and set up redirects on Netlify, so that the source of truth for publication is the `date:` YAML frontmatter, and not the filename.

ericwbailey:
Closed via https://github.com/a11yproject/a11yproject.com/pull/1384.

PR #1384, opened 2021-12-03
Remove dates from post filenames

ericwbailey:
This PR addresses https://github.com/a11yproject/a11yproject.com/issues/1368. It:

- Removes the `yyyy-mm-dd-` prefix from all posts and post image directories.
- Removes references to the `yyyy-mm-dd-` prefix in website content.
- Sets up redirects to not break links.


SaptakS (commented):
Looks good to me. I see the announcements still have the date in their filename, is that on purpose?

ericwbailey:
That would be because I forgot about them 🙃 

Should be fixed now.

SaptakS (approved):
Looks good to me.

Issue #1380, opened 2021-11-21
Upgrade Sass


ericwbailey:
Our current version of Sass is outdated, and its syntax will cause breaking changes when updated. We'll need to:

1. [ ] Update Sass to the latest version.
2. [ ] Verify that Netlify's new VM can process it.
3. [ ] Update all old Sass syntax to use new Sass conventions.

This issue comes from a conversation with @mxmason. We'll also want to reference @esjay's work on https://github.com/a11yproject/a11yproject.com/pull/1141, as this may affect that.

ericwbailey:
Closed via https://github.com/a11yproject/a11yproject.com/pull/1390.

PR #1390, opened 2021-12-13
Upgrade `sass` to `1.45.0`

mxmason:
## Summary
This upgrades our version of `dart-sass` to `1.45.0` and includes cleanup necessary to prevent `stylelint` from crashing during build.

This PR completes 2 of 3 tasks in #1380. We don't need to fully migrate in order to upgrade `dart-sass` (yet).

## Important changes
- `mappy-breakpoints` is now a local dependency in `css/imports`. It's functionally identical to its NPM counterpart, but it makes use of new SASS syntax is so that the project can build. If the published NPM package receives appropriate updates, we can go back to using it, but I don't expect that to happen soon.
- many of our other `.scss` files have been passed through `sass-migrator` so that they utilize new SASS syntax as well.

⚠️ Please verify that nothing important has broken in our design! I don't expect any changes, but still.

### DIFF
--- a/.stylelintrc
+++ b/.stylelintrc
@@ -13,7 +13,7 @@
 		"selector-type-no-unknown": true,
 		"media-feature-name-no-unknown": true,
 		"at-rule-no-unknown": [true, {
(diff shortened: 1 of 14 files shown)


SaptakS (approved):
looks good to me.

PR #1392, opened 2021-12-22
add gaming resources to a11y project

CarlyGerard:
The A11y Project seems like a good spot to gather gaming accessibility resources in one spot, but I didn't see the category yet on the site. So I'm hoping to start the category with a few resources. The resources themselves should be free/low barrier to entry (a Microsoft account is needed for the gaming course I think, but it's free). Happy to adjust if some shouldn't be included.

### DIFF
--- a/src/_data/resources.json
+++ b/src/_data/resources.json
@@ -1927,6 +1927,25 @@
 			]
 		}
 	],
+	"gaming": [
+		{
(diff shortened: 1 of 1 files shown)


ericwbailey:
What a lovely idea. Thank you for adding it!

PR #1402, opened 2022-01-19
Fix features resource links

ericwbailey:
This PR fixes links to our featured resources. The issue was reported on our contact form from two individuals. It also restores our old `slugify` filter, as its absence was causing the announcements partial to fail to render.


ericwbailey:
@mxmason Tagging you for a head's up about the `slugify` bit.

ericwbailey:
A little more info as I dig: it appears Eleventy's slugify filter turned the apostrophe `'` in "Bootcamper's" into a `-` character, while our older filter removed it.

mxmason:
thank you, @ericwbailey; sorry for the mess. i’d like to get some tests written soon to prevent this kind of regression 🙇🏻‍♀️

PR #1423, opened 2022-03-14
Implement base JS and markup for a theme toggle

hankchizljaw:
This addition adds a markup pattern—radio buttons—and some lightweight JavaScript that gets and sets a `user-color-scheme` localStorage value based on their state. 

This state is then applied in two places:

1) As a `data-user-theme` value on the HTML root element 
2) As a `checked` state on the corresponding radio button control

Every time this state changes, the data is saved in localStorage. When a page is loaded with the `apply-theme-setting.min.js` script in the `<head>` *before* the CSS is loaded, the correct theme via localStorage will be applied.

The default theme is system. The intent of that is that the user’s preference will be honoured.

A browser with dev tools showing the radio buttons and the user-color-scheme attribute on the html root. Arrows and boxes demonstrate the relationship.

I've added a temporary page that features the new HTML controls. This is to hopefully make your lives easier when it comes to applying the CSS.

### DIFF
--- a/src/_includes/layouts/base.njk
+++ b/src/_includes/layouts/base.njk
@@ -12,6 +12,10 @@
       media="(prefers-color-scheme: dark)">
 		{% include "meta/social-media.njk" %}
 		{% include "meta/icons.njk" %}
+
+		{# This needs to run before CSS so we don't get a flash of the default theme #}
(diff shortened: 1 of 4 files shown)


ericwbailey (approved):
Lovely stuff, thank you so much!

ericwbailey:
Context for the public record. We commissioned @hankchizljaw to build this functionality, with the added functionality matching our commissioned specifications.

PR #1428, opened 2022-03-18
fix: gaming heading id

jackdclark:
I was just browing the [Resources](https://www.a11yproject.com/resources/) page and realised that the "Gaming" link incorrectly points to the "Books" section. This fixes that.


ericwbailey (approved):
Great catch. Thank you for doing this!

Issue #1426, opened 2022-03-18
Variable-ize colors


nchan0154:
## Overview

I'd like to propose that we migrate from using SASS variables to use CSS Variables instead. This can be a part of larger efforts to rewrite the overall CSS using more modern techniques, but I think this can potentially take priority as a way to set up the ground work for making #1423 viable (but also, not sure how much is in scope for Andy!) and reducing some of our technical debt. 

## Implementation details
Since the site uses different color schemes for each page, we'll likely need several layers of abstractions. I'm using [Hylia](https://github.com/hankchizljaw/hylia/blob/master/src/scss/_theme.scss) and Andy Bell's previous work as a major source of inspiration here. The initial layer should likely be similar to how our colors are currently set up, named colors that correspond to the actual value being described.

```scss
--color-black: #000000;
--color-gray-tint: #e6e6e6;
--color-white: #fff;
```

The second layer should represent the purpose of the color. Our components should only ever reference these kinds of variables, and not directly.
```scss
// Color definitions
--color-card-background: var(--color-white);
--color-card-focus-background: var(--color-light-green);

// Component example
.card {
  background-color: var(--color-card-background);

  &:focus {
      background-color: var(--color-card-focus-background);
  }
}
```

I'm thinking that all definitions for these variables should be set in a single place, if not a single file, at least a single folder where each file represents a particular color theme. I think we can consider seperating themes out from the pages that they represent, so as the site grows, new pages can opt into different color schemes that exist via YAML, rather than us having to copy and paste the variables for a different page into the theme, or [extend current selectors in many different places](https://github.com/a11yproject/a11yproject.com/blob/9e903cec9f392fbda9912f708fe3ca4fc4172dc0/src/css/components/_c-hero.scss).

All this will set us up up for being able to override colors in one place to create dark mode/other color schemes, as we can see in [Hylia.](https://github.com/hankchizljaw/hylia/blob/master/src/scss/_theme.scss)

### Extra considerations
I've thought briefly about the below aspects, but there are likely things that I'm missing, so feel free to weigh in!
- IE11 compatibility. EJ has mentioned to me that we do still want to support IE11. I'm thinking we can either just set a nice, reasonable fallback on each of the components, so that IE11 gets a single color scheme across the whole site, or we use a SASS mixin in order to provide color schemes within IE11. This approach is more fully outlined in [this blog post.](https://vgpena.github.io/winning-with-css-variables/)
[...]

PR #1433, opened 2022-04-09
Add dark mode functionality

nchan0154:
More context and discussion in #1426


SaptakS:
I will try and review this over the weekend.

src/css/components/_c-post.scss
```diff
@@ -196,6 +231,11 @@ $_c-post-vertical-spacing: 4ex;
 	@include link-states($color-black);
 
 	text-underline-offset: 0.2em;
+
+	// User Queries
+	@media (prefers-color-scheme: dark) {
+		@include link-states($color-white);
+	}
```
SaptakS, line 238: I think this and line 231 needs to be updated. Right now it's ending up being `var(--color-#000000)` because that's what $color-black refers to. Probably need a separate override.

src/css/components/_c-banner.scss
```diff
@@ -2,11 +2,11 @@ $_banner-triangle-size: 1.49rem;
 
 
 .c-banner {
-	background-color: $color-green-tint;
-	background-image: url("#{$path-to-img}/resource-banner-inset.svg");
+	@include var(background-color, card-accent);
+	background-image: url("#{$path-to-img}/resource-banner-inset-lightmode.svg");
```
SaptakS, line 6: I think this needs to switch between lightmode and darkmode. Right now it's always in lightmode and so in darkmode it doesn't look the way it should be:

image

src/css/components/_c-person.scss
```diff
@@ -107,11 +107,22 @@ $_person-distance: (
...
+			background-color: $color-black;
+			border-color: $color-gray-tint-darkmode;
+		}
+	}
 }
 
 
 .c-person__link {
-	@include link-states($color-blue-shade);
+	@include link-states($color-blue-shade-lightmode);
```
SaptakS, line 125: $color-blue-shade-lightmode seems to return a color hex and not color name. Hence this results in `--color-#232d71` which gives an error and doesn't render correctly. Can be seen in team page. Same for the dark mode.

src/css/components/_c-person.scss
```diff
@@ -123,6 +134,24 @@ $_person-distance: (
...
+.c-person__spotlight-link {
+	@include link-states($color-blue-shade-lightmode);
+
+	display: inline-block;
+	margin-top: 1rem;
+
+	// User Queries
+	@media (prefers-color-scheme: dark) {
+		@include link-states($color-blue-tint-darkmode);
+	}
```
SaptakS, line 154: Similar to above, $color-blue-shade-lightmode seems to return a color hex and not color name. Hence this results in `--color-#232d71` which gives an error and doesn't render correctly. Can be seen in spotlight page. Same for the dark mode.

SaptakS (changes requested):
This looks great!! I love the override objects and the mixins used to handle the color changes. Also the theming logic with theme color for each page is a good idea. I found a few things that I left comment of.

Few other things I noticed are:
- I think the background color and border color for `<code>` element also needs to be updated based on color mode. Right now in dark mode, the background color for `<code>` still stays light, resulting in insufficient color contrast.
- The color for `.c-post__category__link` is not updated on changing theme, and fails color contrast in dark mode.
- `--color-card-meta-text` needs to be updated for dark mode (I checked in resource pages, which uses green color scheme). In dark mode, it's not passing contrast.
- `.c-contribute__cta` background-color needs to be updated based on changing theme.
- `.c-person__info:hover` background-color needs to be updated based on changing theme. Honestly, I don't think this even needs a color change as such on hover.
- `.c-changelog__version` background-color needs to be updated based on changing theme.
- `.c-last-updated` color doesn't meet contrast check in dark mode, so might need to be updated.

Also, TIL `role="switch"` is not a valid role for `<input type="radio">` based on axe checked and w3 validator. Is there a specific reason we want switch role?

src/_includes/theme-toggle.njk
```diff
@@ -0,0 +1,14 @@
+<section class="l-footer__theme-switcher" aria-label="Theme switcher">
```
SaptakS, line 1: Not specifically a change I am requesting, but seems like this `<section>` element and the parent `<div>` element has the same class. So maybe we want to just keep either the `<section>` or the `<div>` with the aria-label, instead of both? Feels a bit over-structured to me. Is there a specific reason why that might be problematic?

src/_includes/header.njk
```diff
@@ -49,7 +48,7 @@
 	{% if templateClass == "template-spotlight" %}
 		{% set hero = heroes.spotlight %}
 		{% include "heroes/primary-nav.njk" %}
-	{% endif %}
+	{% endif %} #}
```
SaptakS, line 51: I think we can safely remove this comment section fully from the code, since we don't seem to need that anymore.

ericwbailey:
>Also, once this is merged, maybe we can remove [https://github.com/a11yproject/a11yproject.com/blob/main/src/theme-toggle-playground.njk](https://github.com/a11yproject/a11yproject.com/blob/main/src/theme-toggle-playground.njk?rgh-link-date=2022-04-11T17%3A10%3A42Z) ?

Yup! That's the plan.

nchan0154:
@SaptakS Thank you so much for taking the time to review. I had accidentally pushed the remaining few commits to my own personal fork instead of this one here, so you were missing a few of the most recent commits (very much my fault!) but I double checked your list and found things I needed to clean up anyway, so thank you so much for your time and effort!

nchan0154:
Should already be pushed @SaptakS!

SaptakS:
@nchan0154 All the issues mentioned above seems to be fixed. Only one thing I mentioned that is still there is the `role="switch"`. Though it's not a wcag issue, but html validators throw error when having `role="switch"` on `type="radio"`.

ericwbailey:
Let's remove the `role="switch"` info for now. I can revisit the pattern in a subsequent PR, and I think the `fieldset` construct on the radio group will provide sufficient context.

ericwbailey:
For the public record, I'm also _incredibly_ excited about this 🎉

SaptakS (approved):
I think this looks good to be merged.

SaptakS:
@ericwbailey @nchan0154 this is now live!! There was a small bug after removing role=switch which I fixed. Now everything seems to be working great live. In case you wanted to make any announcements, you can now.

ericwbailey:
Woohoo! Thanks to both of you so very much for all your hard work getting this PR merged 🎉

PR #1434, opened 2022-04-18
Uses type=radio in theme toggle JS code

SaptakS:
When I removed role="switch" the JS code for the switching stopped working properly. I have updated the JS code to use `type='radio'` instead of `role='switch'`


ericwbailey:
Thanks for jumping on this, @SaptakS!

PR #1440, opened 2022-05-07
Add Atmos color resource

VojtechVidra:
Add Atmos color resource to resources.json. Atmos is for creating palettes and making sure those colors are accessible at the same time.


ericwbailey:
Thank you for your submission, @VojtechVidra. Unfortunately, we try to only link to web resources like this that are completely free and primarily focused on accessibility.

PR #1456, opened 2022-08-01
Update article how to accessible heading structure

rianrietveld:
This PR is an update for the current info.
- Removes the outdated reference to multiple `H1`s in WHATWG HTML Living standards
- Adds info from the post by Steven Faulkner about how to add `hgroup` in an accessible way.


SaptakS (changes requested):

SaptakS (approved):
Looks good to me.

PR #1468, opened 2022-08-19
Added resource to Courses section

Jfox32199:
Added resource CoursesOnline to Courses section. Website in question has recently begun to list courses that cover various aspects of UX.


ericwbailey:
Thank you for submitting this PR @Jfox32199. Unfortunately, this resource is out of scope of the goals of The A11Y Project. We are a website focused on digital accessibility, and not UX instruction. While UX and accessibility are closely related, this resource is not focused on accessibility.

PR #1469, opened 2022-08-29
Add missing quotation mark on line 24

naiiiden:
`role="<ROLE_TYPE>` 
to 
`role="<ROLE_TYPE>"`


ericwbailey (approved):
Nice catch! Thank you for the fix.

PR #1470, opened 2022-08-31
Add The Neurodivergent Woman Podcast

praesongprasit:
I've been trying to understand more about neurodiverse people, and [The Neurodivergent Woman](https://www.ndwomanpod.com/) was shared at work yesterday.

We don't have any podcast that cover neurodiversity in tech. And unfortunately this isn't one.

It's is as close as it gets though, as they walk us through different conditions and user needs. 

But if we're restricting to tech only, then perfectly happy for it to be excluded :)

### DIFF
--- a/src/_data/resources.json
+++ b/src/_data/resources.json
@@ -1619,6 +1619,13 @@
 			"type": "Episode"
 		},
 		{
+			"title": "The Neurodivergent Woman",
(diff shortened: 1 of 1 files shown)


ericwbailey:
I love it! Thank you for adding it, and congratulations on your first contribution 🎉

PR #1471, opened 2022-08-31
Update Accessible Podcast url

praesongprasit:
The podcast has been retired so accessible.fm url no longer works.

The new suggested link goes to [their archive](https://www.timothybuck.me/accessible-episode-archive).


ericwbailey:
Thanks for the fix!

PR #1551, opened 2023-05-18
Remove the Contact page

ericwbailey:
This PR removes the Contact page. All it really does is collect scams, grifts, backlink schemes, and other unwanted and undesirable messages and I don't want to burden anyone else with that exhausting nonsense.

### DIFF
--- a/src/_includes/nav/footer.njk
+++ b/src/_includes/nav/footer.njk
@@ -63,9 +63,6 @@
 		</h2>
 		<ul class="c-footer__list">
 			<li>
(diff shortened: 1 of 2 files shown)


davatron5000:
@sandrina-p Howdy! The contact page files are actually still in the repo but as you can see were taken down for mental health reasons.

**Here are the files:**

- https://github.com/a11yproject/a11yproject.com/blob/main/src/contact.bak
- https://github.com/a11yproject/a11yproject.com/blob/main/src/js/contact.js

**Here's a CodePen of those all working together:**

- https://codepen.io/davatron5000/pen/poxYxjm

Hope that helps.

PR #1558, opened 2023-05-25
fix(posts): how-to-hide-content - Mention .sr-only

sandrina-p:
Hi team! (cc @davatron5000)

I've updated the post [How-to: Hide content](https://www.a11yproject.com/posts/how-to-hide-content/).

- fix(posts): how-to-hide-content - Mention .sr-only

Why? Personally, instead of `.visually-hidden`, I name the CSS class `.sr-only` because:
- it's shorter and quicker to write and spell.
- It's also popular among CSS libraries, such as Bootstrap, Bulma, Tailwind, (and overall Google search).

For that reason, I think it's important to mention that alternative in your post. What do you think?

Thank you :)

### DIFF
--- a/src/posts/how-to-hide-content.md
+++ b/src/posts/how-to-hide-content.md
@@ -53,7 +53,7 @@ With modern browsers and IE9 and up, the visually hidden selector can be written
 
 This will ensure that if an interactive element receives focus, the styles of the .`visually-hidden` class will be undone and the focusable content will be exposed.
 
-The CSS class `.visually-hidden` is also commonly known as `.sr-only`, where sr stands for "screen readers".
+Many popular CSS frameworks provide a `.visually-hidden` or `.sr-only` class, where `sr` stands for "screen readers".
 
 
 ## Alternatives to `display: none`
--- a/src/posts/how-to-hide-content.md
+++ b/src/posts/how-to-hide-content.md
@@ -41,7 +41,6 @@ The "clip pattern" accomplishes this task for you; hide the content visually, ye
 }
 ```
 
-
 If the `.visually-hidden` class is applied to natively focusable elements (such as `a`, `button`, `input`, etc) they **must** become visible when they receive keyboard focus. Otherwise, a sighted keyboard user would have to try and figure out where their visible focus indicator had gone to.
 
 With modern browsers and IE9 and up, the visually hidden selector can be written like so:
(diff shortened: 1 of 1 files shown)


src/posts/how-to-hide-content.md
```diff
@@ -53,6 +53,8 @@ With modern browsers and IE9 and up, the visually hidden selector can be written
 
 This will ensure that if an interactive element receives focus, the styles of the .`visually-hidden` class will be undone and the focusable content will be exposed.
 
+The CSS class `.visually-hidden` is also commonly known as `.sr-only`, where sr stands for "screen readers".
```
davatron5000, line 56: I think we shy away from the idea that "accessibility is for screen readers only" and that's why we're using `.visually-hidden`. But I think a note about `.sr-only` classes existing in popular in frameworks is good to point out. 

```suggestion
Many popular CSS frameworks provide a `.visually-hidden` or `.sr-only` class, where `sr` stands for "screen readers".
```

davatron5000 (changes requested):
Thanks so much for the improvement. I made a suggestion here, feel free to rewrite but I thought it was important to point out:

- this is provided in some CSS frameworks.
- putting the `sr` by itself in a code block since it's not an English word or acronym

src/posts/how-to-hide-content.md
scottaohara: agreed.  screen readers are merely one type of assistive technology that visually hidden content can leverage or be hindered by.    there was an effort to try to move people away from using/saying "sr-only".  but agreed it is worth mentioning it so at least to inform people they're the same thing.

sandrina-p: I understand that effort and the intentions behind it. 

May I confess that as a non-native English speaker, I struggle to write "visually-hidden" compared to sr-only. Also, I've seen people confusing it with CSS `visibility: hidden` because of its similar name.
- visibility-hidden -> more inclusive
- sr-only -> easier to use

With this said, do you know of any other possible naming (even if not common) for this technique that could be both easy and inclusive?

scottaohara: i've seen 'at-only' as another option.  i don't know how well that caught on though... i wouldn't necessarily mention it.  but yah, naming things is hard, particularly when one needs to be both descriptive with the name, while concise and aware of everyone's level of understanding/comfort with the name.   :: giant everlasting shrug ::

sandrina-p: Interesting name but agree that it shouldn’t be mentioned. Thanks @scottaohara , I have nothing else to say here

davatron5000 (approved):

davatron5000:
🚀🎉 👏   Thanks @sandrina-p

PR #1560, opened 2023-05-31
adds post about common UI behaviors that break color blindness

HarcourtHamsa:


HarcourtHamsa:
@davatron5000 can you look into this at your spare time?

PR #1566, opened 2023-07-05
Move no-js class to html element

szepeviktor:
Just that.

From https://darekkay.com/evaluatory/demo/6.html


davatron5000:
Thanks for catching this. In general, please provide more context around your fixes than "Just that." The link to Dare Kay's took was provided without context. 

- [x] `class` cannot be applied to the `<!doctype html>` declaration
- [x] I looked through the project and there's no use of `.no-js` in script or CSS

 I almost rejected this PR for lack of context

szepeviktor:
Thank you for merging it!

PR #1607, opened 2025-01-02
Remove broken link to Invision in sponsors.json

jlbooker:
Invision has shutdown, so this link is broken.


ericwbailey (approved):
Thanks for doing this. It's a bummer, but probably best to remove.

PR #1628, opened 2025-07-25
Add GetWCAG and Update resources.json

dellaak:
Add getwcag


ericwbailey:
https://www.a11yproject.com/posts/should-i-use-an-accessibility-overlay/

PR #1655, opened 2026-05-15
Shorten sponsor link labels and expose taglines via aria-describedby

KatieSeeMePls:
## Summary

Each sponsor logo in the sidebar is wrapped in a link, and the image's `alt` text serves as both the picture description and the link's accessible name. That works well for the shorter entries, but three of the seven (A11Y Collective, Be Inclusive, Hello Access) currently run 25 to 30 words. As link labels those strings are verbose, and can become a real annoyance for screen reader users, since the whole sentence is announced every time the link is encountered, including in the links rotor where users are usually trying to scan a flat list quickly. The Hello Access alt also does not lead with the brand name, so the sponsor cannot be identified without listening through the full sentence.

This change splits the two roles the alt was playing, so each sponsor has a short, scannable link label and a longer descriptive tagline that screen reader users can choose to hear in context.

## What changed

**`sponsors.json`** now has three fields per sponsor in place of `alt`:

- `linkLabel` becomes the image `alt` and therefore the link's accessible name, e.g. `"Fable, sponsor of The A11Y Project"`.
- `tagline` is the longer descriptive copy. Rendered into a visually hidden `<p>` next to the link.
- `slug` is a stable identifier used to wire the two together.

**`ad.njk`** renders both pieces and connects them via `aria-describedby`. The Fable entry has no tagline, so the template gracefully falls back to a label-only link (no `aria-describedby`, no empty `<p>`).

Redundant `role="img"` attribute removed from each `` since image elements already have that role implicitly.

## Why split the label from the description

`aria-describedby` lets the same content live in two channels of the screen reader experience:

- Users skimming the links rotor hear just the short label and can scan past quickly.
- Users who land on a specific link get the tagline announced after a brief pause.
- Users with verbose descriptions turned off can skip the tagline entirely.

Sighted users see exactly the same logos and marketing pitches as before; the pitch already lives inside each logo image as pixels.

## Test plan

- [ ] Visual: sponsor cards render identically in light and dark themes across mobile, tablet, and desktop breakpoints
- [ ] Screen reader (NVDA, JAWS, VoiceOver): navigate to sponsor section, confirm link reads as short label; description follows after a pause
- [ ] Links rotor: open the flat links list, confirm sponsor links show short labels only
- [ ] Fable card (no tagline): renders without `aria-describedby` and without an empty `<p>`
- [ ] Keyboard: tab through sponsor links; focus order and indication unchanged


KatieSeeMePls:
Apologies for jumping straight to a PR. I should have filed an issue first per the contributing guidelines. Happy to close this and open an issue for discussion if that's preferred, or to keep it open as a working reference if you'd like to weigh in on the approach. No pressure to merge.

PR #1665, opened 2026-06-20
Add AxeGuard resource

Reevie:
Adds AxeGuard to the Development tools resources list.

AxeGuard is a free online WCAG scanner that uses axe-core in a real browser. It reports automated findings only and does not claim compliance or replace manual accessibility review.


ericwbailey:
https://github.com/a11yproject/a11yproject.com/issues/1667

